PR #2, opened 2019-05-11
support for headings

lpfennigschmidt:
- adds first implementation of block parser to recognise atx-headings
- also changes fontSize in Editor when typing in a heading
(please have a look at the tests or MarkdownHeading class >> attributesFor: for changing font sizes and tell me if you know a better way :D)


jonaskordt (changes requested):
For a first version I think the parsing structure is fine.

But using the class object of the MarkdownHeading class in the way it is used is not fine in my opinion. Having the matches: method in the class is fine. Here it makes sense not to instantiate an object because we don't even know if we will need one. But having #attributesFor: and #levelOf: in the class is very weird to me. It would make much more sense to instantiate a MarkdownHeading object and then let it implement #attributes and #level. (It would obviously need to know its content for this but that would be much more consistent with the inline parser too.)

packages/MarkdownEditor-Core.package/MarkdownHeading.class/class/attributesFor..st
```diff
@@ -0,0 +1,11 @@
+text attributes
+attributesFor: aString
+	
+	| headingLevel attributes |
+	headingLevel := self levelOf: aString.
+	attributes := OrderedCollection new.
+	((headingLevel + 1) isDivisibleBy: 2) ifTrue: [attributes add: TextEmphasis underlined].
+	(headingLevel between: 1 and: 2) ifTrue: [attributes add: (TextFontReference toFont: (TextStyle default copy fontOfSize: 26))].
```
felix-gohla: The following lines (or at least the font sizes) could be extracted into a method containing a constant. :blush:

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/testEmptyHeading.st
```diff
@@ -0,0 +1,7 @@
+tests
+testEmptyHeading
+
+	self
+		assert: '#'
+		parsesToRuns: {1}
```
felix-gohla, line 6: Nice idea! 👍

packages/MarkdownEditor-Core.package/MarkdownBlockParser.class/instance/parse..st
```diff
@@ -0,0 +1,13 @@
+parsing
+parse: aString
+
+	| styledText |
+	styledText := aString asText.
+	aString lineIndicesDo: [:start :endWithoutDelimiters :end | | line |
+		line := aString copyFrom: start to: end.
+		(MarkdownHeading matches: line)
+			ifTrue: [
+				(MarkdownHeading attributesFor: line) do: [:attribute |
```
felix-gohla: Why line break before?

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/assert.parsesToRuns.values..st
```diff
@@ -0,0 +1,9 @@
+accessing
+assert: markdownString parsesToRuns: runs values: values
+
+	| parsedText |
+	parsedText := parser parse: markdownString.
+	self
+		assert: markdownString equals: parsedText string;
```
felix-gohla, line 7: Why does this need to be ensured? For now, this is correct, but the method doesn't express this check in the title.

felix-gohla (changes requested):
It's so cool that its working this responsive. Very nice! ☺️

Would be nice when some logic wouldn't be on class site. I think this would get incomprehensible when more blocks will appear in the class hierarchy. Furthermore, the `attributesFor:` method could be decomposed a little.

Nice to see those special rules with leading spaces work!

packages/MarkdownEditor-Core.package/MarkdownBlockParser.class/instance/parse..st
pixunil: The method isn't settled yet, so I think formatting will come later

packages/MarkdownEditor-Core.package/MarkdownHeading.class/class/attributesFor..st
lpfennigschmidt: Yep, noted :D

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/assert.parsesToRuns.values..st
pixunil: I couldn't think of a better name. As our parser returns a `Text`, we want to test against this text, which is made up of a `string` and `runs` which has another time `runs` and `values`. I want to keep all three asserts. But I'll gladly collect naming proposals.

felix-gohla (approved):
Looks very clean to me now!

jonaskordt (approved):
Great improvement!

kiragrammel (approved):
Great :)

PR #4, opened 2019-05-19
Feature/parser emphasis

jonaskordt:


packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/cutFromEnd..st
```diff
@@ -0,0 +1,11 @@
+attributes
+cutFromEnd: aLength
+
+	| cutDelimiter |
+	cutDelimiter := self copy
+							content: (self content last: aLength);
+							position: self endPosition - aLength + 1;
+							yourself.
+	self content: (self content copyFrom: 1 to:  self endPosition - aLength).
```
jonaskordt: This line needs to be changed. Using `self endPosition` doesn't make sense. We want a copy of the string except for the last `aLength` characters but `endPosition` is the position of the end of the token in the whole string that is being parsed. Instead it would make sense to use `self content first: self length - aLength`. This gives us the first `self length - aLength` characters of the content.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/isRightFlankingFor..st
```diff
@@ -0,0 +1,21 @@
+testing-private
+isRightFlankingFor: aMarkdownReadStream
+
+	"1 not preceded by whitespace AND"
+	"   2a not preceded by punct OR"
+	"	2b preceded  by punct and followed by ws or punct"
```
jonaskordt: should we keep this comment?

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/interpretTokens.st
```diff
@@ -0,0 +1,30 @@
...
+						ifTrue: [| cut | 
+							cut := (currentToken cutFromStart: length) asStringToken.
+							self tokens 
+								add: cut 
+								before: currentToken.
+							currentToken := cut]
+						ifFalse: [currentToken := self tokens at: (self tokens indexOf: currentToken) put: currentToken asStringToken]]].
+	currentToken := self tokens 
+		after: currentToken
+		ifAbsent: [^ self]]
```
jonaskordt: This is the method which I think needs to be refactored somehow. Currently I'm not sure yet how. If there are any questions about it please ask. I know it's not quite easy to understand (which is why it needs refactoring...). 

One thing I can tell you is that we don't use the built in squeak iterating over collections because we are modifying the collection while iterating over it. Because of that we need to treat it like a doubly linked list. We can do that by having a pointer to the `currentToken` we are looking at and then using `after: currentToken` to get the next one and using `add:after:` and `add:before:` to insert into the collection. 

One more thing that makes it even more complicated is that we sometimes have to change the class of the `currentToken`. For this we obviously need a new object which means the `currentToken` needs to be replaced in the collection and also we need to change our pointer to it (line 27) because the old object is no longer in the collection so we couldn't use `after:` to get the next one. 

In addition to that we sometimes don't want to look at the next token but instead look at the same token again (when a token is longer that 2 * we need to cut off the first 2 * and then look at the rest again to see if we can find another opener for it). In this case we simply set the `currentToken` pointer to the StringToken we insert before the `currentToken` anyways (line 26) so that `after:` will give is the previous `currentToken` once again.

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/initialize.st
```diff
@@ -0,0 +1,6 @@
+initialize-release
+initialize
+
+	super initialize.
+	self tokenClasses: {MarkdownDelimiterEmphasis. MarkdownString.} asSortedCollection.
+	self tokenClasses sortBlock: [:a :b | a precedence > b precedence]
```
pixunil, line 8: I think it is too early to use a precedence mechanism at this stage (this line is currently a no-op).
If you are absolutely sure that you want to use this mechanism, then keep it, otherwise I'm fine with removing it.

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/tokenClasses..st
```diff
@@ -0,0 +1,4 @@
+accessing-private
```
pixunil, line 1: How do you differentiate between `accessing` and `accessing-private`? Currently, the non-private methods are only called in the tests which is fine for private methods too.

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/tokenize..st
```diff
@@ -0,0 +1,10 @@
+parsing
+tokenize: aString
+
+	| mdStream |
```
pixunil: `markdownStream` (no abbreviations)

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/initialize.st
```diff
@@ -0,0 +1,6 @@
+initialize-release
+initialize
+
+	super initialize.
+	self tokenClasses: {MarkdownDelimiterEmphasis. MarkdownString.} asSortedCollection.
```
pixunil: No `.` after the last element of an array

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/nextTokenIn..st
```diff
@@ -0,0 +1,7 @@
+parsing
+nextTokenIn: aMarkdownReadStream
+
+	| tokenClass |
+	tokenClass := self tokenClasses detect: [:c | 
+		c beginsWith: aMarkdownReadStream peek asString].
```
pixunil: Is there a reason why you compare with a `String`? The emphasis tokens work with `$* asString` and the catch-all tokens always `^ true`. So, either ask `c` with the stream (something like `c accepts: aMarkdownReadStream` and do the peeking in `c` as needed), or use `Character`, as you only peek one step.

Also, `c` could get a nicer name, but I can't come up with something good, as `tokenClass` shadows the method variable.

packages/MarkdownEditor-Core.package/MarkdownToken.class/class/newFor..st
```diff
@@ -0,0 +1,7 @@
+instance creation
+newFor: aMarkdownReadStream
+
+	^ self basicNew
```
pixunil: I think `basicNew` is unnecessary, as you don't need to circumvent something in `initialize`. To make things worse, you use in `asStringToken` a normal `new`, which negated the status as initializer of `initialize: stream`

I prefer a setter for the stream, e.g. `fillContentFrom:`.

packages/MarkdownEditor-Core.package/MarkdownToken.class/instance/position.st
```diff
@@ -0,0 +1,4 @@
+accessing
+position
```
pixunil: `startPosition` would match `endPosition` nicely

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/initialize..st
```diff
@@ -0,0 +1,12 @@
+initialize-release
+initialize: aMarkdownReadStream
+
+	| len |
+	super initialize.
+	len := self calculateLengthFor: aMarkdownReadStream.
+	self content: (aMarkdownReadStream peek: len);
```
pixunil: You seam to count the asterisks only to peek `len` times ahead to create your content. Could `calculateLengthFor:` be rewritten to `consumeContentFor:` which does this in one go?

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/isLeftFlankingFor..st
```diff
@@ -0,0 +1,17 @@
+testing-private
+isLeftFlankingFor: aMarkdownReadStream
```
pixunil: `decideLeftFlankingFor:` would match `decideFlankingFor:` as it is more near to it than to `isLeftFlanking`.
Also, this doesn't feel like `testing` to me, more like `initialize-release` or `deciding`.
You could also move the setter in here, but I'm indifferent.

Same for `isRightFlankingFor:`.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/isFollowedByAnyOf.in..st
```diff
@@ -0,0 +1,6 @@
+testing-private
+isFollowedByAnyOf: aCollectionOfCharacters in: aMarkdownReadStream
```
pixunil: Should move into `MarkdownReadStream`. You seam to have the constraint that the stream is in the state of being exactly at the delimiter, hence the `-private`, but it should be clearer when the sender directly acts on the stream.

Same for `isPrecededByAnyOf:`.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/leftFlanking..st
```diff
@@ -0,0 +1,4 @@
+accessing-private
+leftFlanking: aBoolean
```
pixunil: Hmm, I don't know what guidelines are saying, but it feels strange that this is named differently.
Couldn't setter and instance variable be named `isLeftFlanking`?

Same for `isRightFlanking:`.

packages/MarkdownEditor-Core.package/MarkdownAttribute.class/instance/end..st
```diff
@@ -0,0 +1,4 @@
+accessing
+end: anInteger
```
pixunil: `startPosition` / `endPosition` like in Tokens? Or are they different here?

packages/MarkdownEditor-Core.package/MarkdownReadStream.class/methodProperties.json
```diff
@@ -0,0 +1,7 @@
+{
+	"class" : {
```
pixunil, line 2: I'm sorry, I can't comment on the correct file; didn't you want to add a class comment to `MarkdownReadStream` something like "This is for compatibility" (and for stuff we always wanted to have)?

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/interpretTokens.st
pixunil: Some other problems I'm seeing here that there are many operations which could be done by the Tokens or Attributes. For instance, you ask a closer "Here is your opener, would you please generate an attribute from it?" And later, "Hi opener, how long are you?" and either "Looks like you are consumed. Please make yourself comfortable in your second life as StringTocken" or "Oh, you have so much potential, but I only need `length` characters from you. Please cut yourself a bit".
(that good dark very quick...)
I have not fully thought about it, but maybe you could do lots of the coordination in the tokens themselves. Like, the opener cuts itself (even to zero) and returns a remainder (either the unmatched thing, or the string token). They already have the knowledge how many characters are needed (btw; you ask the attribute and generate magic numbers -> polymorphism).

Or, you could let the tokens remember how many characters were consumed.

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/parse..st
```diff
@@ -0,0 +1,7 @@
+parsing
+parse: aString
+
+	self tokenize: aString.
```
pixunil: I think it is a bit weird to have the tokens as a instance variable (with the attributes, you could make more operations possible, but tokens are only used once). I'd suggest to make them a temp variable here.

packages/MarkdownEditor-Core.package/MarkdownAttribute.class/instance/squeakTextAttributes.st
```diff
@@ -0,0 +1,4 @@
+accessing
+squeakTextAttributes
```
pixunil: just `textAttribute`, it makes me wonder whether there is `SqueakTextAttribute`

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/class/punctuationChars.st
```diff
@@ -0,0 +1,5 @@
+constants
+punctuationChars
+
+	"ASCII punctuation characters according to GFM spec. :)"
+	^ #($! $" $# $$ $% $& $' $( $) $* $+ $, $- $. $/ $: $; $< $= $> $? $@ $[ $] $\ $^ $_ $` ${ $} $| $~)
```
pixunil: This is just a reminder to ask next meeting: `Unicode` has `categoryOf:` which would make this obsolete (and would have the power of unicode). Sadly, it returns the category, like `Pf`, is encoded as a magic number.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
```diff
@@ -0,0 +1,10 @@
+attributes
+attributeStartingAt: aMarkdownDelimiterEmphasis
```
pixunil: I found this method hard to understand, as it puts the receiver into the role of a closer and the arguments as opener. Maybe it would help to reverse this to `attributeTo: closingMarkdownDelimiterEmphasis`. Nevertheless, a semantic-typed arg name would help too

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
```diff
@@ -0,0 +1,10 @@
...
+
+	| attrLen start end attributeClass |
+
+	attrLen := {self length. aMarkdownDelimiterEmphasis length. self boldLength} min.
+	start := aMarkdownDelimiterEmphasis endPosition - attrLen + 1.
```
pixunil: Also, this index shifting is driving me crazy... Are attributes inclusive and tokens not or are things just like this is written?

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/isRightFlankingFor..st
pixunil: It says what is written in code, so remove it. You could add a reference to the spec, but I don't think it would help. Also, you could make use of `==>` in the return expression (but if you think that matching wordings matter more, that's fine too)

pixunil (changes requested):
My first bunch of comments (I won't review the tests today), I hope they help for your refactoring.
Some names I'm proposing aren't ideal, feel free to think about a better one :)
I can follow the flow of parsing rather good, but there are some discrepancies in naming and responsibilities. Also, sometimes there is a index shifting which I can't really understand.

packages/MarkdownEditor-Core.package/MarkdownAttribute.class/instance/end..st
jonaskordt: Here it is basically the start position of the opening token and the end position of the closing token

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
jonaskordt: As far as I can remember everything is inclusive. It's simply a bit complicated because we want the tokens to be included as bold or italic. But if we have an opener that is 5 stars long and it opens a bold section then we only want its last two stars to be a part of that bold section. For that reason we need to turn an inclusive ending to an inclusive start which requires a +1 sadly...

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/isFollowedByAnyOf.in..st
jonaskordt: We have it here because the Delimiter knows its length. I really think it makes sense like this

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/leftFlanking..st
jonaskordt: I don't understand what you mean :D

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/parse..st
jonaskordt: My thinking was that basically the whole inline Parser is one huge method object. Thus we want most of the variables to be instance so that we don't have to keep tossing them around as parameters. But in other parts of the class we haven't done this which makes this part here look weird. I would like the make it more into a method object and introduce even more instance variables which would make decomposing some of the methods much easier

packages/MarkdownEditor-Core.package/MarkdownAttribute.class/instance/end..st
pixunil: It was meant as request to rename both instance variables. But rereading it, it could read as a question. When they both more or less have the same meaning, I think that `startPosition` and `endPosition` would be less confusing

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/leftFlanking..st
pixunil: I mean it is weird that the getter is `isLeftFlanking` (which I like), but instance variable and setter have the shorter form `leftFlanking`. Like I said, I have no idea what guidelines are saying, when in doubt, ask at our next meeting

jonaskordt: Well usually for boolean setter you would use `beLeftFlanking` and `beNotLeftFlanking` but then you would need both and that's kind of weird. Not sure why the best option is

packages/MarkdownEditor-Core.package/MarkdownAttribute.class/instance/end..st
felix-gohla: Think so, too. Renaming would be a little bit more precise.

packages/MarkdownEditor-Core.package/MarkdownAttribute.class/instance/squeakTextAttributes.st
felix-gohla: :+1:

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/class/precedence.st
```diff
@@ -0,0 +1,4 @@
+accessing
+precedence
+
+	^ 42
```
felix-gohla: Yeah... I think we should probably refactor this whole precedence thing into a chain of responsibility. For now, magic numbers were a good start, but I don't want to see this in production.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/class/punctuationChars.st
felix-gohla: Well, `Unicode` does not necessarily conform to the GFM. In the spec, these are (nearly) all the characters being mentioned as punctuation! `Unicode` on the other hand has more characters classified as punctuation.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
felix-gohla: I would suggest `attributeUpTo: closingMarkdownDelimiterEmphasis`.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
```diff
@@ -0,0 +1,10 @@
+attributes
+attributeStartingAt: aMarkdownDelimiterEmphasis
+
+	| attrLen start end attributeClass |
+
+	attrLen := {self length. aMarkdownDelimiterEmphasis length. self boldLength} min.
```
felix-gohla: I don't like the name `attrLen` in here. This is no attribute. We should rename it to something like `delimiterPartLen`.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
felix-gohla: We would need to change both `position` and `endPosition` in order to fix this. Or find a better name for `attrLen` that expresses that it's some kind of offset.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/calculateLengthFor..st
```diff
@@ -0,0 +1,11 @@
+initialize-release
+calculateLengthFor: aMarkdownReadStream
+
+	| streamCopy count |
+	aMarkdownReadStream atEnd ifTrue: [Error signal: 'Stream end reached'].
+	streamCopy := aMarkdownReadStream copy.
+	count := 0.
+	[streamCopy next = self class asterisk] 
+		whileTrue: [ count := count + 1].
```
felix-gohla: This bracket style. 😆

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/cutFromEnd..st
felix-gohla: Agreed. 👍

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/initialize..st
felix-gohla: We could probably do that, maybe this would be great for all of the other tokens which will come up. E.g. I personally don't like the fact that the string token only contains one char. It would be great to see other tokens determine their length, too.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/initialize..st
```diff
@@ -0,0 +1,12 @@
...
+		decideFlankingFor: aMarkdownReadStream
+
```
felix-gohla: Oh please remove those lines.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/isFollowedByAnyOf.in..st
felix-gohla: @pixunil How would you express the fact that we need a offset to the current stream position in the function name?

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/isLeftFlankingFor..st
felix-gohla: Agreed.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/isRightFlankingFor..st
felix-gohla: Please remove it.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/matches..st
```diff
@@ -0,0 +1,9 @@
+testing
+matches: aMarkdownDelimiterEmphasis
+
+	 ^ (self isOpenerAndCloser or: [aMarkdownDelimiterEmphasis isOpenerAndCloser])
```
felix-gohla: We could insert a reference to the spec here. Finding this method the first time, I think you would have trouble understanding its sense.

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/initialize.st
felix-gohla: Like I said some comments above, I would be happy refactoring it into a chain of responsibility.

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/interpretTokens.st
```diff
@@ -0,0 +1,30 @@
...
+	| currentToken |
+	self attributes: OrderedCollection new.
+	currentToken := self tokens first.
+	[true] whileTrue: [(currentToken isDelimiter and: [currentToken isCloser]) 
+		ifTrue: [self findOpenerFor: currentToken 
+				ifFound: [:opener | | attribute length |
+					attribute := currentToken attributeStartingAt: opener.
+					self attributes add: attribute.
+					attribute isBold 
+						ifTrue: [length := 2]
```
felix-gohla: `currentToken boldLength`?!

felix-gohla: @jonaskordt You remember the collection minimum? This is already in the tokens and would express exactly the same thing.

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/interpretTokens.st
felix-gohla: I argued with Jonas in the sprint last week about exactly this topic. It would make this cutting mechanism more... intuitive, I guess? We should talk to @kiragrammel and ask her for her opinion.

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/nextTokenIn..st
felix-gohla: The reason are the upcoming `![` tokens. We need to somehow make sure, we can peek more than one char. Giving the stream to the `MarkdownToken` is a little counterintuitive, I think. Don't have a clue how this could be done nicely yet.

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/parse..st
felix-gohla: Passing them around as `aCollectionOfMarkdownTokens` would make the lines look like giant deformities. 😃 

@jonaskordt Which parts do you mean?

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/tokenClasses..st
felix-gohla: I don't know why, but I think `attributes:`, `tokens:` and `tokens` should go into `accessing-private`, to make clear they shouldn't be touched when using the inline parser. `accessing-private` is for all those accessors that should only be used by the object itself during operation.

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/tokenize..st
felix-gohla: 👍

packages/MarkdownEditor-Core.package/MarkdownReadStream.class/methodProperties.json
felix-gohla: We should probably add class comments to all of the classes. 🙃

packages/MarkdownEditor-Core.package/MarkdownToken.class/class/newFor..st
felix-gohla: It doesn't feel like initializing anymore using a setter-method. But this needs to be discussed.

packages/MarkdownEditor-Core.package/MarkdownToken.class/class/newFor..st
```diff
@@ -0,0 +1,7 @@
+instance creation
+newFor: aMarkdownReadStream
+
+	^ self basicNew 
+		initialize: aMarkdownReadStream;
+		position: aMarkdownReadStream position + 1;
```
felix-gohla: No request for change, but I hate the way Squeak is handling streams and positions. 🙄😂

packages/MarkdownEditor-Tests.package/MarkdownDelimiterEmphasisTest.class/instance/testNotRightFlanking.st
```diff
@@ -0,0 +1,12 @@
+tests - state
+testNotRightFlanking
+
+	| stream delimiter |
+	"Don't use it that way. Content is set just for testing."
+	delimiter := MarkdownDelimiterEmphasis new content: '***'; yourself.
+	
+	stream := self readStreamOn: 'abc *** def' at: 4.
+	self deny: (delimiter isRightFlankingFor: stream).
+	stream := self readStreamOn: ' ***'.
```
felix-gohla: Please insert newline before.

packages/MarkdownEditor-Tests.package/MarkdownInlineParserTest.class/instance/testInterpretTokensBold.st
```diff
@@ -0,0 +1,16 @@
+tests - interpretation
+testInterpretTokensBold
+
+	| tokens |
+	tokens := OrderedCollection new.
+	tokens add: (MarkdownDelimiterEmphasis new content: '**'; position: 1; leftFlanking: true; rightFlanking: false; yourself).
+	tokens add: (MarkdownString new content: 'f'; position: 3; yourself).
```
felix-gohla: Why do we set a `position` on the string/char tokens? We don't need them like shown in `testDelimiterCantCloseItself`.

packages/MarkdownEditor-Tests.package/MarkdownInlineParserTest.class/instance/testInterpretTokensCutDelimiterClosesTwice.st
```diff
@@ -0,0 +1,20 @@
...
+	tokens add: (MarkdownDelimiterEmphasis new content: '*'; position: 1; leftFlanking: true; rightFlanking: false; yourself).
+	tokens add: (MarkdownString new content: 'a'; position: 2; yourself).
+	tokens add: (MarkdownDelimiterEmphasis new content: '**'; position: 3; leftFlanking: true; rightFlanking: true; yourself).
+	tokens add: (MarkdownString new content: 'b'; position: 5; yourself).
+	tokens add: (MarkdownDelimiterEmphasis new content: '***'; position: 6; leftFlanking: false; rightFlanking: true; yourself).
+	inlineParser tokens: tokens.
+	inlineParser interpretTokens.
+	self assert: 2 equals: inlineParser attributes size.
+	self assert: (inlineParser attributes first isKindOf: MarkdownAttributeBold).
+	self assert: (inlineParser attributes second isKindOf: MarkdownAttributeItalic).
```
felix-gohla, line 14: Please move down right before line 18.

felix-gohla (changes requested):
Overall, this is pretty decent. We could make a demo branch for testing it and showing it to Stephan. Well done, team! 🥇

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
jonaskordt: we use it to determine the length of the attribute we are creating though. I think it makes sense. I would just like to remove the abbreviation

felix-gohla: We use it to determine the length, but it is not the length. It is just the length of both delimiters that will get cut off.

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/parse..st
jonaskordt: for example we could split up the `interpretTokens` method in smaller methods if we created an instance variable for `currentToken` and maybe some more

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
jonaskordt: Oh yea you're right. I misread. We basically determine the length of the part of the Delimiter which is responsible for the attribute. SO something like `attributeDelimiterLength` would describe it, but that's kind of long...

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/interpretTokens.st
jonaskordt: I agree this would be better but I don't quite know how exactly to do it.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
lpfennigschmidt: Also, I stumbled over the class/object name in this method. The object is actually still a delimiter, yet 'DelimiterEmphasis' puts emphasis at the end so it reads like its an emphasis for something. It would be clearer if it was called MarkdownDelimiterForEmphasis or MarkdownEmphasisDelimiter

packages/MarkdownEditor-Core.package/MarkdownReadStream.class/instance/last.st
```diff
@@ -0,0 +1,5 @@
+accessing
+last
+
+	(self position = 0) ifTrue: [^ Character lf].
+	^ super last
```
lpfennigschmidt, line 5: Maybe we should add a comment on why we overwrite certain methods of the normal ReadStream for our purposes

packages/MarkdownEditor-Core.package/MarkdownReadStream.class/instance/next..st
```diff
@@ -0,0 +1,11 @@
...
+next: anInteger
+
+	| ans endPosition |
+
+	endPosition := position + anInteger  min:  readLimit.
+	ans := collection copyFrom: position+1 to: endPosition.
+	(ans size ~= anInteger) 
+		ifTrue: [ans := ans padded: #right to: anInteger with: Character space].
+	position := endPosition.
+	^ans
```
lpfennigschmidt: a better name for 'ans' could be 'nextCharacters'

packages/MarkdownEditor-Core.package/MarkdownReadStream.class/instance/next..st
```diff
@@ -0,0 +1,11 @@
+accessing
+next: anInteger
+
+	| ans endPosition |
+
+	endPosition := position + anInteger  min:  readLimit.
+	ans := collection copyFrom: position+1 to: endPosition.
+	(ans size ~= anInteger) 
+		ifTrue: [ans := ans padded: #right to: anInteger with: Character space].
```
lpfennigschmidt: I'm still not sure about padding the result with spaces, do we have a use case for that or is it actually just a way for not throwing exceptions in that situation?

packages/MarkdownEditor-Core.package/MarkdownString.class/class/beginsWith..st
```diff
@@ -0,0 +1,4 @@
+testing
+beginsWith: aString
```
lpfennigschmidt: The wording confuses me, maybe 'mayBeginWith' makes it clearer that we don't have a MarkdownString yet, we just want to find out whether we can have one like this
Same for MarkdownDelimiterEmphasis

packages/MarkdownEditor-Tests.package/MarkdownDelimiterEmphasisTest.class/instance/testEmphasisAttributeBold2.st
```diff
@@ -0,0 +1,17 @@
+tests - attributes
+testEmphasisAttributeBold2
```
lpfennigschmidt: maybe 'testDifferentDelimitersCreateBoldAttribute' and then rename the above test case to 'testSameDelimitersCreateBoldAttribute'

packages/MarkdownEditor-Tests.package/MarkdownDelimiterEmphasisTest.class/instance/testIsCloser.st
```diff
@@ -0,0 +1,17 @@
+tests - state
+testIsCloser
```
lpfennigschmidt: Maybe 'testClosingDelimiterIsCloser' and 'testNotClosingDelimiterIsNotCloser', only using the method name as the name for the test case can be quite confusing (as it is in this case, cause like, is the test coming closer or is it further away? xD)

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/isFollowedByAnyOf.in..st
pixunil: Huh, the more I'm thinking about it, the more I want to move stuff into the reader (it is rather passive, but could be more active). I'd say to leave this as it is atm and look later how much we make use of the stream

packages/MarkdownEditor-Tests.package/MarkdownDelimiterEmphasisTest.class/instance/testMatchesOpener.st
```diff
@@ -0,0 +1,14 @@
+tests - matching
+testMatchesOpener
```
lpfennigschmidt: renaming the test cases to their assert-statements would help realising what you're testing, so here that would be 'testOpenerAndCloserMatchesOpener'

lpfennigschmidt (changes requested):
my first thoughts, I'm gonna comment some more later on

packages/MarkdownEditor-Core.package/MarkdownReadStream.class/instance/next..st
jonaskordt: Line ends and beginnings are supposed to be seen as spaces. Thats why we do it.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/initialize..st
jonaskordt: We decided to keep it like it is for now, because we don't see the value of changing it right now.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/leftFlanking..st
jonaskordt: We decided to keep it

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/class/punctuationChars.st
jonaskordt: We decided to keep it like this and add a link to the spec

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
jonaskordt: We decided not to reverse it but to change the parameter name to `openingMarkdownDelimiterEmphasis`

jonaskordt: We don't understand. Please elaborate

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/interpretTokens.st
jonaskordt: We changed quite a lot. Check it out :)

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
lpfennigschmidt: The MarkdownDelimiterEmphasis-Object is supposed to be a delimiter rather than an emphasis, right? Like, if I want to imagine the object, its a delimiter with a content (* or ** or whatever) and is handled as a token for this content. Its effect afterwards is the text emphasis, but the object itself is not an emphasis. So I mean, I would consider renaming the whole class into MarkdownEmphasisDelimiter or MarkdownDelimiterForEmphasis, because just having DelimiterEmphasis makes it hard to understand what the object is supposed to be or to do

packages/MarkdownEditor-Core.package/MarkdownReadStream.class/instance/next..st
lpfennigschmidt: Yeah, but we still get 5 spaces, if I peek for the next 8 characters and there are only 3 left in the stream, right? I'm not sure whether we should have that, but I guess it does not matter for now if we do. It just might be a point of confusion for other people, if we do not necessarily need it but they might think we do

packages/MarkdownEditor-Core.package/MarkdownAttribute.class/instance/isBold.st
```diff
@@ -0,0 +1,4 @@
+accessing
+isBold
```
lpfennigschmidt: there are no senders of this method

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/class/beginsWith..st
```diff
@@ -0,0 +1,4 @@
+testing
+beginsWith: aString
```
lpfennigschmidt: this method is deprecated now, right?

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeDelimiterLengthWith..st
```diff
@@ -0,0 +1,4 @@
+attributes
+attributeDelimiterLengthWith: aMarkdownDelimiterEmphasis
```
lpfennigschmidt: this name is very hard to understand. What you are actually trying to do is whether to find out, if a delimiter creates italic or bold attributes with another delimiter, right? So could you also return the correct attribute at this point or do you explicitly need the number 1 or 2?

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
```diff
@@ -0,0 +1,10 @@
+attributes
+attributeStartingAt: openingMarkdownDelimiterEmphasis
+
+	| attributeDelimiterLength start end attributeClass |
+
+	attributeDelimiterLength := (self attributeDelimiterLengthWith: openingMarkdownDelimiterEmphasis).
```
lpfennigschmidt: could attributeDelimiterLength just be delimiterLength? The attribute-part is a little confusing cause the name doesn't specify what its doing with or for the attribute

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/cutOpener.to..st
```diff
@@ -0,0 +1,10 @@
+parsing
+cutOpener: openingMarkdownDelimiterEmphasis to: aLength
+
+	openingMarkdownDelimiterEmphasis length > aLength 
+		ifTrue: [self tokens
```
lpfennigschmidt: in cutCloser: to: you return this whole statement, I'm guessing it should be similar in both methods. Although you seem not to need the return value from this, maybe its better to still return it for consistency

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/interpretTokens.st
```diff
@@ -0,0 +1,16 @@
...
+	[true] whileTrue: [(currentToken isDelimiter and: [currentToken isCloser]) 
+		ifTrue: [self findOpenerFor: currentToken 
+				ifFound: [:opener |  | length |
+					self attributes add: (currentToken attributeStartingAt: opener).
+					length := currentToken attributeDelimiterLengthWith: opener.
+					self cutOpener: opener to: length.
+					currentToken := self cutCloser: currentToken to: length]].
+		currentToken := self tokens 
+			after: currentToken
+			ifAbsent: [^ self]]
```
lpfennigschmidt: Good job shortening this method ; )

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/tokenize.st
```diff
@@ -0,0 +1,10 @@
+parsing
+tokenize
+
+	| markdownStream |
+	self tokens: OrderedCollection new.
+	markdownStream := MarkdownReadStream on: self text asString.
+	[markdownStream atEnd] whileFalse: [| token |
+		token := self nextTokenIn: markdownStream.
+		self tokens addLast: token.
```
lpfennigschmidt, line 8: You could also use self tokens add: token, because add: uses addLast:, but if you want to explicitly say where it gets added, you can keep it

packages/MarkdownEditor-Core.package/MarkdownString.class/class/beginsWith..st
lpfennigschmidt: Also deprecated :D

packages/MarkdownEditor-Tests.package/MarkdownDelimiterEmphasisTest.class/instance/testBeginsWith.st
```diff
@@ -0,0 +1,7 @@
+tests - recognition
+testBeginsWith
+
+	self assert: (MarkdownDelimiterEmphasis mayBeginWith: '*').
+	
+	{'-'. '!'. '!['. '['. 'f'. ''. ' '} do: [:string |
+		self deny: (MarkdownDelimiterEmphasis beginsWith: string)]
```
lpfennigschmidt: you're still using beginsWith: here

packages/MarkdownEditor-Tests.package/MarkdownDelimiterEmphasisTest.class/instance/testBeginsWith.st
```diff
@@ -0,0 +1,7 @@
+tests - recognition
+testBeginsWith
```
lpfennigschmidt: maybe put this into two test cases, one being testBeginsWithAsterisk and the other testDoesNotBeginWithOtherCharacter

packages/MarkdownEditor-Tests.package/MarkdownDelimiterEmphasisTest.class/instance/testDifferentDelimiterLengthsAttributeDelimiterLength.st
```diff
@@ -0,0 +1,11 @@
+tests - matching
+testDifferentDelimiterLengthsAttributeDelimiterLength
```
lpfennigschmidt: What? xD 
What is this test case supposed to tell me that it is testing?

packages/MarkdownEditor-Tests.package/MarkdownDelimiterEmphasisTest.class/instance/testMatchesOpenerNot.st
```diff
@@ -0,0 +1,17 @@
+tests - matching
+testMatchesOpenerNot
```
lpfennigschmidt: the assert statements in this test and in testOpenerAndCloserMatchesOpener are the same only one time its asserted and the other is denied, so the different setup should be in the test case name. This one here does not match, because both delimiters' lengths are not divisible by 3, right? And in the other test case they do match because they are both 3? Or do they not match in this test case, because their length is different?

packages/MarkdownEditor-Tests.package/MarkdownDelimiterEmphasisTest.class/instance/testSameDelimiterCreateItalicAttribute.st
```diff
@@ -0,0 +1,17 @@
+tests - attributes
+testSameDelimiterCreateItalicAttribute
```
lpfennigschmidt: for consistency testSameDelimiters with plural s? :D

packages/MarkdownEditor-Core.package/MarkdownReadStream.class/instance/next..st
pixunil: I'd like to see a quote from the specs where it does say that eof should be treated like spaces. Afaik eol or beginning of line should be treated like whitespace which is specific to emphasis and can be implemented in the stream, but I still have vague feelings for it

packages/MarkdownEditor-Tests.package/MarkdownInlineParserTest.class/instance/testAddAttributesToText.st
```diff
@@ -0,0 +1,2 @@
+tests - attributes
+testAddAttributesToText
```
lpfennigschmidt: empty test? :D

packages/MarkdownEditor-Tests.package/MarkdownStringTest.class/instance/testInitialization.st
```diff
@@ -0,0 +1,12 @@
+tests
+testInitialization
+
+	| mdString stream |
```
lpfennigschmidt: markdownString :D

packages/MarkdownEditor-Tests.package/MarkdownStringTest.class/instance/testPosition.st
```diff
@@ -0,0 +1,8 @@
+tests
+testPosition
+
+	| stream mdString |
```
lpfennigschmidt: markdownString

packages/MarkdownEditor-Tests.package/MarkdownTokenTest.class/instance/testDelimeterEmphasisAsString.st
```diff
@@ -0,0 +1,11 @@
+tests - conversion
+testDelimeterEmphasisAsString
```
lpfennigschmidt: found a typo xD

lpfennigschmidt (changes requested):
Nicely done, especially on that interpretTokens-method! :D
I just have some naming things. I know some are very tedious, maybe we can come to an agreement, that changes like writing markdownString rather than mdString do not have to be commented and can be changed and committed by someone else (if that works for PR's, I hope so :D)

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
```diff
@@ -0,0 +1,10 @@
...
+	start := openingMarkdownDelimiterEmphasis endPosition - attributeDelimiterLength + 1.
+	end := self startPosition + attributeDelimiterLength - 1.
+	attributeClass := self attributeTypeStartingAt: openingMarkdownDelimiterEmphasis.
```
pixunil: Isn't the type dependant on the `attributeDelimiterLength`? The method body of `attributeTypeStartingAt:` could be simpler when using only the length imo

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/decideFlankingFor..st
```diff
@@ -0,0 +1,5 @@
+initialize-release
+decideFlankingFor: aMarkdownReadStream
+
+	self leftFlanking: (self decideLeftFlankingFor: aMarkdownReadStream).
+	self rightFlanking: (self decideRightFlankingFor: aMarkdownReadStream).
```
pixunil: Cascade

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/class/punctuationChars.st
```diff
@@ -0,0 +1,5 @@
+constants
+punctuationChars
```
pixunil: Should move to instance side (as it is only used there atm).
Would also consider renaming it to `punctuationCharacters`.
Same for `whitespaceChars`.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
```diff
@@ -0,0 +1,10 @@
+attributes
+attributeStartingAt: openingMarkdownDelimiterEmphasis
+
+	| attributeDelimiterLength start end attributeClass |
+
```
pixunil: Unnecessary newline

packages/MarkdownEditor-Tests.package/MarkdownInlineParserTest.class/instance/testFindOpenerForEmphasisDelimiter.st
```diff
@@ -0,0 +1,30 @@
...
+	inlineParser tokens: tokens.
+	
+	inlineParser findOpenerFor: closer 
+		ifFound: [:o | foundOpener := o].
+	self assert: opener equals: foundOpener.
+	
+	tokens remove: opener.
+	inlineParser tokens: tokens.
+	inlineParser findOpenerFor: closer 
+		ifFound: [:o | self fail]
```
pixunil, line 29: Idea:
This is a separate test case. You modify the `tokens` collection so I had to keep the state mentally. I prefer to have one test with `{opener. string. closer}` and one with `{string. closer}`.
Also, when the first assert is moved in its own method, you can write `ifFound: [^ self assert: ...]. self fail`

packages/MarkdownEditor-Tests.package/MarkdownStringTest.class/instance/testBeginsWith.st
```diff
@@ -0,0 +1,5 @@
+tests
+testBeginsWith
```
pixunil: `testCanBeginWithAnyCharacter`

packages/MarkdownEditor-Tests.package/MarkdownInlineParserTest.class/instance/testInterpretTokensItalics.st
```diff
@@ -0,0 +1,16 @@
...
+	tokens add: (MarkdownString new content: 'f'; startPosition: 2; yourself).
+	tokens add: (MarkdownString new content: 'o'; startPosition: 3; yourself).
+	tokens add: (MarkdownString new content: 'o'; startPosition: 4; yourself).
+	tokens add: (MarkdownDelimiterEmphasis new content: '*'; startPosition: 5; leftFlanking: false; rightFlanking: true; yourself).
+	inlineParser tokens: tokens.
+	inlineParser interpretTokens.
+	self assert: 1 equals: inlineParser attributes size.
+	self assert: (inlineParser attributes first isKindOf: MarkdownAttributeItalic).
+	self assert: 1 equals: inlineParser attributes first startPosition.
+	self assert: 5 equals: inlineParser attributes first endPosition
```
pixunil: Question:
Could we use `=` to compare things like `MarkdownAttribute`? There are four lines which looks like you are decomposing an object into its variables (which make it somewhat hard to read as you always fetch the object from your uut four times)

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/class/whitespaceChars.st
```diff
@@ -0,0 +1,5 @@
+constants
+whitespaceChars
+
+	"Whitespace characters according to GFM spec: https://github.github.com/gfm/#whitespace-character"
```
pixunil: _ASCII_ whitespace characters

packages/MarkdownEditor-Tests.package/MarkdownDelimiterEmphasisTest.class/instance/testNotLeftFlanking.st
```diff
@@ -0,0 +1,12 @@
+tests - state
+testNotLeftFlanking
```
pixunil: Idea: this could be decomposed into `testIsNotLeftFlankingAtEndOfLine`, `testIsNotLeftFlankingWhenFollowedWithWhitespace`, ...

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/matches..st
```diff
@@ -0,0 +1,10 @@
+testing
+matches: aMarkdownDelimiterEmphasis
```
pixunil: Discuss:
`matches` is very generic. Maybe we could use `isMatchingOpenerFor: aClosingDelimiterEmphasis`?

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/fillContentFrom..st
```diff
@@ -0,0 +1,7 @@
+initialize-release
+fillContentFrom: aMarkdownReadStream
+
+	| len |
```
pixunil: `length`

pixunil (changes requested):
Only a few requests left, mostly code style.
Some comments are annotated with "Idea" or "Question". Those don't need to be addressed in this PR.
You might noticed that I unresolved some resolved comments. Those were meant to be discussed with the whole team, so they should be left open until then.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
jonaskordt: I think once we have more Delimiter Types it will make sense to be honest. Also I really want Delimiter to come first in the name because it we have MarkdownEmphasisDelimiter and MarkdownLinkDelimiter it would be weird to have Delimiter at the end. Also I don't think having a For in the name would make a difference. It kind of clutters the name. Thus I would like it to stay like it is.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeDelimiterLengthWith..st
jonaskordt: We actually need the numbers 1 and 2 because they determine how much we need to cut off the delimiters.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
jonaskordt: We actually discussed this name a bit. What we are determining is the length of the part of the delimiter that belongs to the attribute (it's either 1 or 2). We thought about having `Part` in the name swell but we thought it would be too long. Thats why we settled for `attributeDelimiterLength`.

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/tokenize.st
jonaskordt: I think it's a good idea to explicitly say that we are adding it last because if it wouldn't be added last then the whole thing could break

packages/MarkdownEditor-Tests.package/MarkdownDelimiterEmphasisTest.class/instance/testDifferentDelimiterLengthsAttributeDelimiterLength.st
jonaskordt: Please suggest a better name We couldn't find a better one.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
jonaskordt: Do I understand correctly that you want the method `attributeTypeStartingAt:` to call `attributeDelimiterLengthWith:`?

pixunil: No, I thought of a method which maps `1` to italic and `2` to bold (which is more explicit how you are doing the mapping). The variable `attributeDelimiterLength` already contains the length, so why not use it?

packages/MarkdownEditor-Core.package/MarkdownReadStream.class/instance/next..st
felix-gohla: You're right, both EOL and Beginning Of Line shall be treated as Unicode Whitespace:
[GFM Spec Flanking](https://github.github.com/gfm/#left-flanking-delimiter-run)
> For purposes of this definition, the beginning and the end of the line count as Unicode whitespace.

Because as of now only the `MarkdownInlineParser` uses the `MarkdownReadStream` and because it shall only be passed text line by line (or multiple lines being part of exactly one paragraph), we decided to put the method into the stream.

pixunil (approved):
Most of the stuff I would be commenting are higher-order refactoring, which I have already left a bunch of comments. And I'm too tired to make more naming proposals...
Approval under the condition that https://github.com/hpi-swa-teaching/MarkdownEditor/pull/4#discussion_r286206211 will be addressed.

felix-gohla (approved):
I am happy with the state. After renaming to `MarkdownEmphasisDelimiter`, this can be merged into master. 👍

packages/MarkdownEditor-Core.package/MarkdownReadStream.class/instance/next..st
jonaskordt: We changed it so that one space is added but if `next:` is called with an integer that would require adding more than one space an Error is raised.

packages/MarkdownEditor-Core.package/MarkdownDelimiterEmphasis.class/instance/attributeStartingAt..st
jonaskordt: We could rename it to `delimiterLength` as soon as we rename `MarkdownEmphasisDelimiter` to `MarkdownEmphasisDelimiterRun`. But we want to discuss this renaming with everyone

lpfennigschmidt (approved):

kiragrammel (approved):
Good work! Now it looks tidier and is easier to understand.

felix-gohla (approved):
👍

PR #8, opened 2019-05-28
Paragraph Block

pixunil:
Previously, we created one block-like object for each line.
Now, paragraphs can be extended or interrupted with blank lines, whitespace lines or headings.


packages/MarkdownEditor-Core.package/MarkdownBlockParser.class/instance/createBlockFor..st
```diff
@@ -1,6 +1,6 @@
 parsing
-findBlockFor: aString
+createBlockFor: aString
```
pixunil: We forgot to delete this method...

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/assert.parsesToBlocks..st
```diff
@@ -0,0 +1,9 @@
+accessing
+assert: aCollectionOfStrings parsesToBlocks: aCollectionOfMarkdownBlockClasses
```
pixunil: This was a idea to have multiple lines without ugly line breaks in the code. But if you think that this method is unneccessary overhead, we'll use the simpler solution.

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testAsTextStylesInline.st
```diff
@@ -0,0 +1,12 @@
+tests
+testAsTextStylesInline
```
pixunil: I hope this method is understandable. I tried to make clear that the mock inline parser just returns itself to check if the parser is used with the expected inline content, but if this is not clear, we'll think of something else.

packages/MarkdownEditor-Tests.package/MarkdownParagraphTest.class/instance/blankLine.st
```diff
@@ -0,0 +1,5 @@
+constants
+blankLine
```
felix-gohla, line 2: Why no newline char?

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/assert.parsesToBlocks..st
jonaskordt: I like it except for the fact that the second argument is the expected collection. Usually in Squeak the first argument is the expectation.

Suggestion: `assert: aCollectionOfMarkdownBlockClasses isParsedStructureOf: aCollectionOfLines`

felix-gohla: Why don't you use `'text', Character linefeed, 'text in new line'`? The below tests are quite hard to understand this way. They don't contain the concept that there are newlines in between the strings.

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/testHeadingFollowedByParagraphWithInlineEmphasis.st
```diff
@@ -0,0 +1,8 @@
+tests - text with inline
+testHeadingFollowedByParagraphWithInlineEmphasis
+
+	self
```
felix-gohla, line 4: Many of these tests are kinda redundant, are they? The runs are checked somewhere else.

felix-gohla (commented):
Didn't read all the tests, yet. But the part I've read until now looks good.

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testAsTextStylesInline.st
jonaskordt: I really don't think this is understandable. I don't like how we are using the MockObject. It is not clear why `assert: inlineParserMock identical: result` should be the case.

Also It needs to be tested that parsing the inline content doesn't remove the Heading Styling. But maybe that should be in a different test for the inline parser.

jonaskordt (changes requested):

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/testHeadingFollowedByParagraphWithInlineEmphasis.st
pixunil: Which of redundancy do you mean?
The tests under `tests - text with inline` are meant as integration tests (they get quite complicated though), so they of course test already covered features by unit tests.

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/assert.parsesToBlocks..st
pixunil: If we would use a "line collection" throughout our project, it would enable us a more compact and understandable definition.
Yes, it hides that there are newlines, but using `'# Heading', Character cr, 'Paragraph'` loses the focus on the content to the newlines imo. (afaik `Character cr` is more broadly used then `Character linefeed`)

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/testHeadingFollowedByParagraphWithInlineEmphasis.st
pixunil: The tests in `tests - text with inline` are meant as acceptance test.
That said, they will test more complex situations which should already be covered by unit tests.

jonaskordt: I think in general we should ask somebody about redundancy in integration and acceptance tests. It's pretty common in our tests by now but I personally think that it makes sense because the integration and acceptance tests give you a better idea of how a class is used than having to go through all the tiny unit tests. As I said this is only my opinion. I don't know what the best practice is, but I think we should find out and then follow the best practice.

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/testHeadingWithFollowingLineDoesNotContinueEmphasis.st
```diff
@@ -0,0 +1,8 @@
+tests - text with inline
+testHeadingWithFollowingLineDoesNotContinueEmphasis
+
+	self
+		assert: '# Title **not
```
felix-gohla: `Character cr`? Personally, I think the newlines in the code look weird.

packages/MarkdownEditor-Tests.package/MarkdownParagraphTest.class/instance/testBlankLineIsExtendableWithBlankLine.st
```diff
@@ -0,0 +1,6 @@
+tests - expandability
```
felix-gohla: In your tests you use the word `Extendable` instead of `Expandable`. I know, the meaning of the words are quite the same but I think, `tests - extendability` is the better name for the category.

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/assert.parsesToBlocks..st
felix-gohla: Then maybe we should consider renaming the convenience test method to make it a little more clear.

felix-gohla (changes requested):

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testAsTextStylesInline.st
lpfennigschmidt: Yep, I agree, I don't see what you're trying to do just from reading the code. And maybe the name could be something like 'testConvertingHeadingToTextAppliesInlineStyles' or something like this, if that is what it does

lpfennigschmidt (changes requested):

kiragrammel (approved):
In general the code looks good. I just like to talk about the newlines in the tests (like Felix said they look ugly) and the inlineParserMock. 
Maybe you can write class comments.

lpfennigschmidt (approved):

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testConvertingToTextUsesInlineParser.st
```diff
@@ -0,0 +1,12 @@
...
+testConvertingToTextUsesInlineParser
+
+	| markdownString mockInlineParser heading result |
+	markdownString := '# **Stronger** heading with *emphasis*'.
+	mockInlineParser := MarkdownMockInlineParser new.
+	heading := MarkdownHeading newFrom: markdownString.
+	heading inlineParser: mockInlineParser.
+	result := heading asText.
+	self
+		assert: mockInlineParser markerText equals: result;
```
jonaskordt: If I understand correctly here we only want to test if the `mockInlineParser` as received the message `parse:`. Why do we do this by comparing two strings? Couldn't the `mockInlineParser` have a Boolean variable that starts as `false` and gets set to `true` once it has received the message `parse:`? Then we could ask it `hasBeenAskedToParse` or something like that and then this line could be replaced with `self assert: mockInlineParser hasBeenAskedToParse`. 

Or even simpler: Line 12 already asserts that the `mockInlineParser ` remembers the String. It can only remember the String if it has been asked to parse the String. Thus why do we even have to test if it has been asked to parse when we also test that it memorizes the correct String, which it only can if it has been asked to parse it?

jonaskordt (changes requested):
In general I like the `Character cr` thing. But it does kind of bloat the tests. So one suggestion I have would be to give the TestCase subclasses that use it an instance variable called `cr` and initialize it to `Character cr` and then use the instance variable in the tests. Thats just a suggestion though not a request for changes.

The other comment I left is much more important in my opinion because like it is now I still had to take way too much time to understand the test.

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/assert.isParsedStructureOf..st
```diff
@@ -0,0 +1,7 @@
+accessing
+assert: aCollectionOfMarkdownBlockClasses isParsedStructureOf: markdownString
```
felix-gohla, line 2: I like 👍

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testConvertingToTextUsesInlineParser.st
felix-gohla: +1 for the Boolean version.

packages/MarkdownEditor-Tests.package/MarkdownInlineParserTest.class/instance/testPlainParsingKeepsAttributes.st
```diff
@@ -0,0 +1,7 @@
+tests - parsing
+testPlainParsingKeepsAttributes
+
+	| text parsedText |
+	text := Text string: 'I have attributes from my caller' attribute: TextEmphasis underlined.
+	parsedText := inlineParser parse: text.
+	self assert: parsedText hasRuns: {32} values: {{TextEmphasis underlined}}
```
felix-gohla: Please format like `testEmphasisParsingKeepsAttributes`.

packages/MarkdownEditor-Tests.package/MarkdownParagraphTest.class/instance/testConvertingToTextUsesInlineParser.st
```diff
@@ -0,0 +1,12 @@
+tests
+testConvertingToTextUsesInlineParser
+
+	| markdownString mockInlineParser paragraph result |
+	markdownString := 'We may include', Character cr, '*emphasis* or', Character cr, '**strong emphasis** in paragraphs'.
+	mockInlineParser := MarkdownMockInlineParser new.
+	paragraph := MarkdownParagraph newFrom: markdownString.
+	paragraph inlineParser: mockInlineParser.
+	result := paragraph asText.
+	self
```
felix-gohla, line 10: Same problem as in `MarkdownHeadingsTest`.

packages/MarkdownEditor-Tests.package/MarkdownParagraphTest.class/instance/testConvertingToTextUsesInlineParser.st
```diff
@@ -0,0 +1,12 @@
...
+		assert: mockInlineParser markerText equals: result;
```
felix-gohla: I really don't understand this.

felix-gohla (approved):
Thanks, guys. It's really good!

Only thing I don't understand is the `MarkdownMockInlineParser`. Maybe you could explain the sense of it and the `markerText`.

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testConvertingToTextUsesInlineParser.st
pixunil: Actually, I intented to test two things:

1. The result of the inline parser is actually returned (as in, the block parser not just runs it and then forgets about the results)
2. The inline parser was fed with the correct string

The second assert is easier to understand and can be implemented in multiple ways (which is requested by you). The first assert, if the result is actually returned untouched is done by mocking the inline parsing as if it would always return a magic value, `markerText`. As we receive the `markerText`, we can be sure that the inline parsing was passed down to the text-converting.

Imo this makes the test valuable, because the block parser is asserted to follow the correct path of parsing. So, I see two options: We further improve the test, then I'll need more information, what you don't understand, where a comment or other code structures would be helpful. We remove the first assert, because it adds only little value for you (the option I obviously would not prefer).

Regarding your off-topic request: No, `cr` as instance variable is not a good idea, because it makes the string un-inspectable (at least in my two-minute class). Yes, there is bloat, and I would be happy to see a more compact solution with good semantics.

jonaskordt: Okay I understand now. 

The problem is, that it is not clear that the `mockInlineParser` always returns the marker String. Thus one doesn't understand that after line 9 the `result` should be the marker string of the `mockInlineParser` and because of that line 11 doesn't make sense.

I think it would be easier to understand if one could set the string that is always returned by the `MockInlineParser`. Then that would have to happen in the test using a method like `parseResult: aText` and then you could assure that `result` equals `aText`. 

This would eliminate the markerString which is very confusing. Also you would see that the `MockInlineParser` has a fixed parsing result which would make it easy to understand why `result` should be equal to the `parseResult`.

pixunil: > The problem is, that it is not clear that the `mockInlineParser` always returns the marker String.

This should (now) be pretty obvious after looking into the source of the mock object (It belongs to the test case, so I think I can expect the reader to know how it works. It's really simple, so this should be not a big problem).

I'd prefer to rename the name of `markerText` to `parsingResult` or similiar. Your proposed change looks like you have a problem with the semantics of the `markerText` which also can be addressed by renaming the variable instead of micro-managing the mock object

jonaskordt: Renaming the `markerText` to `parsingResult` would already be a huge improvement. The only problem I have with that would be that the String which is returned by `parsingResult` comes out of nowhere. I personally would say that that is weird and kind of ugly but since that is such a small  detail it won't hinder my approval. Maybe the others (@lpfennigschmidt, @kiragrammel, @felix-gohla) could comment which version they would prefer.

jonaskordt: Renamed to `parsingResult` for now. We can discuss this further in the future

jonaskordt (approved):

PR #11, opened 2019-06-10
Fixes theme changing bug

kiragrammel:
- adds reset method to MarkdownTextStyler


kiragrammel:
Should we write tests for `privateFormat:` and `privateStyle:`?

packages/MarkdownEditor-Tests.package/MarkdownTextStylerTest.class/instance/testApplyingThemeDoesNotThrowError.st
```diff
@@ -0,0 +1,9 @@
+as yet unclassified
+testApplyingThemeDoesNotThrowError
+
+	| textMorph |
+	textMorph := PluggableTextMorphPlus new
```
felix-gohla: Maybe this could go into `MarkdownEditorTests`? The `applyUserInterfaceTheme` method does not exist in Squeak 5.0. `MarkdownEditor open` would then be unnecessary.

```suggestion
	MarkdownEditor open.
        self shouldnt: [CommunityTheme new apply] raise: Error
```

packages/MarkdownEditor-Core.package/MarkdownTextStyler.class/instance/reset.st
```diff
@@ -0,0 +1,4 @@
+private
+reset
+
+	"Needed for changing theme. We have nothing to reset."
```
felix-gohla, line 4: 🤣

felix-gohla (changes requested):
Well done :D But please fix for Squeak 5.0.

packages/MarkdownEditor-Tests.package/MarkdownTextStylerTest.class/instance/testApplyingThemeDoesNotThrowError.st
felix-gohla: Ah no worries, I didn't see Squeak 5.0 doesn't even have themes... 😞

pixunil: Actually, I find it hard to understand that you are effectively deactivating this case for 5.0. I think it wuld be clearer with something like
```smalltalk
(#('Squeak5.0') includes: Smalltalk version) ifTrue: [^ self]
```
instead of guarding the assert with a complicated `ifFalse:`.
I think that we shouldn't do `shouldnt:raise: Error` stuff, the test would become red either way, right?
Also, the actual bug occured when Themes are switched; could this be expressed directly?
I would also propose that this is a test for `MarkdownEditor`, there you also have a similiar setup.

Sorry for the long comment, it is more test-theoretical nature as I also don't know which is a proper way.

packages/MarkdownEditor-Core.package/MarkdownTextStyler.class/instance/reset.st
jonaskordt: glorious 😅

jonaskordt (approved):

pixunil (changes requested):

felix-gohla (approved):
When we asked the Hirschfeld people... I'm fine with it. :blush:

packages/MarkdownEditor-Tests.package/MarkdownEditorTests.class/instance/testApplyingThemeDoesNotThrowError.st
```diff
@@ -0,0 +1,10 @@
+tests
+testApplyingThemeDoesNotThrowError
+
+	| textMorph |
+	textMorph := PluggableTextMorphPlus new
```
pixunil: You have a `toolbuilder` to build a `TextMorph` like it is in our application:
https://github.com/hpi-swa-teaching/MarkdownEditor/blob/e1053bf3baa21b38b6fcbf1a045f40704b388d99/packages/MarkdownEditor-Tests.package/MarkdownEditorTests.class/instance/testTextPersistence.st#L5

pixunil (approved):
Cool, it looks more cleaner to deactivate the test case in `expectedFailures`.
If you have the time to do it, you could simplify it a bit more, by using the `toolbuilder` (note to self: link common code lines instead of writing gliberish).
Also, you could rename `MarkdownEditorTests` to `MarkdowEditorTest` (for consistency).
But that's just a nice to have, you can also merge and do this later

jonaskordt (approved):

felix-gohla (approved):
I would like to merge this into the master because we need it for refactoring the parsers into text stylers. :blush:

PR #12, opened 2019-06-10
Fenced Code Blocks

pixunil:
Adds interruption mechanism for block parser.

### Ideas
During development, we had some ideas which should be discussed in our next meeting.

- Add a `Line` object which can be asked after indentation, does some matching etc.
- Make `InlineParser` a `TextStyler` so that blocks choose a styler and a region where it is applied. With these information, `asText` could be implemented in `MarkdownBlock`

### 5.0 Weirdness
Regex is not available in 5.0, so we added the package to the baseline. Confusingly, the tests of the regex package will run along ours. To complete the weirdness, their package fails with >= 5.1, so we would get fails... Thus, we have the `ifTrue:` hack in our baseline.


jonaskordt (approved):
I'm very happy with the code itself. 

I think we should talk either to Stephan or Patrick or both about the Baseline thing but for now I'm okay with it. Maybe there is some way to include the package but not have all the tests running or something.

kiragrammel (approved):
Nice!

packages/MarkdownEditor-Tests.package/MarkdownCodeBlockTest.class/instance/testIndentation1Closes.st
```diff
@@ -0,0 +1,7 @@
+tests - extendability
+testIndentation1Closes
```
lpfennigschmidt: Maybe these three or four tests can be refactored into one, as we discussed yesterday

packages/MarkdownEditor-Tests.package/MarkdownCodeBlockTest.class/instance/testMatch1Indentation.st
```diff
@@ -0,0 +1,4 @@
+tests - matching
+testMatch1Indentation
```
lpfennigschmidt: same refactoring possibility here ; )

packages/MarkdownEditor-Core.package/MarkdownCodeBlock.class/class/matches..st
```diff
@@ -0,0 +1,5 @@
+testing
+matches: aString
+
+	^ (aString beginsWith: '    ') not
+		and: [aString matchesRegex: ' *```+[^`]*']
```
lpfennigschmidt, line 5: for someone who is not familiar with how these regexes work, its a bit hard to understand, why this one matches to backticks with an info string, but not to backticks with an info string with backticks, but I doubt a comment would do any more good here. Maybe you have an idea but if you don't, you can also just leave it like this.

lpfennigschmidt (approved):
Nice job, guys :D
Now we have beautiful coloured smalltalk text 🎉

packages/MarkdownEditor-Core.package/MarkdownCodeBlock.class/class/matches..st
pixunil: Like discussed face-to-face, we should always favour a simple solution over a regex. But regarding how the method looked before, a regex would is more understandable.
(And we can assume that people know the `[^a]` syntax)

packages/BaselineOfMarkdownEditor.package/BaselineOfMarkdownEditor.class/instance/baseline..st
```diff
@@ -5,7 +5,12 @@ baseline: spec
 		for: #'common'
 		do: [
 			spec
-				package: 'MarkdownEditor-Core';
+				package: 'VB-Regex' with: [
+					spec repository: 'http://www.squeaksource.com/Regex'].
+			spec
+				package: 'MarkdownEditor-Core' with: [
+					(#('Squeak5.0') includes: Smalltalk version)
```
pixunil, line 12: (Question migrated from Telegram)
As we told smalltalkCI to only include our tests here:
https://github.com/hpi-swa-teaching/MarkdownEditor/blob/4cb5903f40b34b17f11993cbbe405929c3a15f65/.smalltalk.ston#L14

Do we still need the `#('Squeak5.0') includes: Smalltalk version`?

Otherwise, is this ready to be merged?

felix-gohla:
Rerequested review for refactored tests.

packages/BaselineOfMarkdownEditor.package/BaselineOfMarkdownEditor.class/instance/baseline..st
felix-gohla: Well, it works, but overwriting the post 5.1 Regex would be fatal in my opinion.

pixunil: As far as I understand baselines, you define dependencies here (similiar to a package manager). Then how would we overwrite the `VB-Regex` package when we just say "we are depending on it"?
After a quick search, AutoTDD, Vivide, and Widgets do also have the Regex package in their baseline; AutoTDD without version switch, the other with. I remember that we asked how to isolate their tests from ours, but not whether we can remove the version switch. So I guess we have to ask again...

packages/MarkdownEditor-Core.package/MarkdownCodeBlock.class/instance/asText.st
```diff
@@ -0,0 +1,9 @@
+converting
+asText
+
+	| code result |
+	code := self smalltalkStyler styledTextFor: self codeContent asText.
```
pixunil: After reading the `STTextstyler` class comment, `styledTextFor:` is wrong here. It also does formatting (which could be dangerous if it actually modifies the text) whereas `style:` has a slight different interface but should be correct.
I tried to make a quick test but it got late and I think this issue is complex enough for a pair.
Or, if you don't care about the test, we can just replace this line ^^

Rough STR:

1. Open _Preference Browser_
2. Enable `syntaxHighlightingAsYouTypeLeftArrowAssignment`
3. Open _Markdown Editor_ and type a code block, including `:=`
  _Your image should be frozen_

packages/BaselineOfMarkdownEditor.package/BaselineOfMarkdownEditor.class/instance/baseline..st
felix-gohla: Yeah I know, these should be dependencies. But when they were only dependencies, why should the tests have been failed for 5.1+? This means the package must have been loaded into the image (despite the fact, the dependency is already fulfilled), don't they?

packages/MarkdownEditor-Core.package/MarkdownCodeBlock.class/instance/asText.st
felix-gohla: We should definitely fix this! Regression test would be nice (as discussed face to face).

felix-gohla (changes requested):

packages/MarkdownEditor-Core.package/MarkdownCodeBlock.class/instance/asText.st
pixunil: The solution involved the use of `privateStyle:` as `style:` styles a _copy_ of the string and then informing its view. As we have no adequate view (and we found it strange to make a code block a view), we just call `privateStyle:`...

lpfennigschmidt (approved):
Looks good to me :D
You still want to add a regression test for that image freezing thing?

felix-gohla:
> Looks good to me :D
> You still want to add a regression test for that image freezing thing?

No, no test. We would test behaviour that is not in our system anymore.

packages/MarkdownEditor-Core.package/MarkdownCodeBlock.class/instance/asText.st
```diff
@@ -2,7 +2,8 @@ converting
 asText
 
 	| code result |
-	code := self smalltalkStyler styledTextFor: self codeContent asText.
+	code := self codeContent asText.
+	self smalltalkStyler privateStyle: code.
```
jonaskordt, line 6: I think it's weird to use a private method so maybe we should ask someone if this is how it's supposed to be used.

jonaskordt (approved):

kiragrammel (approved):

felix-gohla (approved):

packages/MarkdownEditor-Core.package/MarkdownCodeBlock.class/instance/asText.st
pixunil: It's approved officially, but should get a comment

packages/BaselineOfMarkdownEditor.package/BaselineOfMarkdownEditor.class/instance/baseline..st
pixunil: You are correct, `VB-Regex` is in 5.1+ a patched version which we would overwrite. Leaving it as it is

PR #14, opened 2019-06-13
Feature/truetype fonts

lpfennigschmidt:
Implements headings using Truetype-fonts


jonaskordt (approved):
Having the `headingFontSized: pointSize` twice is kind of weird but I don't have a better solution right now.

packages/MarkdownEditor-Core.package/MarkdownHeading.class/instance/fontSizes.st
```diff
@@ -0,0 +1,4 @@
+accessing
+fontSizes
+	
+	^ Dictionary newFrom: {1->34. 2->28. 3->22. 4->19. 5->16. 6->14}
```
pixunil, line 4: Wouldn't be an `OrderedCollection` sufficient?

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/testHeadingFollowedByParagraphWithInlineEmphasis.st
```diff
@@ -4,4 +4,8 @@ testHeadingFollowedByParagraphWithInlineEmphasis
 	self
 		assert: '# Title **bold**', Character cr, 'Line *italic*'
 		parsesToRuns: {8. 8. 1. 5. 8}
-		values: {{TextFontReference toFont: (TextStyle default copy fontOfSize: 26). TextEmphasis underlined}. {TextFontReference toFont: (TextStyle default copy fontOfSize: 26). TextEmphasis underlined. TextEmphasis bold.}. {TextFontReference toFont: (TextStyle default copy fontOfSize: 26). TextEmphasis underlined}. {}. {TextEmphasis italic}}
\ No newline at end of file
+		values: {{TextFontReference toFont: (self headingFontSized: 34)}. 
+				{TextFontReference toFont: (self headingFontSized: 34). TextEmphasis bold.}.
```
pixunil: Unnecessary `.`
```suggestion
				{TextFontReference toFont: (self headingFontSized: 34). TextEmphasis bold}.
```

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/headingFontSized..st
```diff
@@ -0,0 +1,7 @@
+accessing
+headingFontSized: pointSize
```
pixunil, line 2: Couldn't you create a `TextFontReference` here? All calls will wrap the font into it, which adds some bloat to the code imo

pixunil (changes requested):
I don't understand the fix for the fonts, but that's not our problem.
Either way, its nice to see that this adds only small changes to our code!

packages/MarkdownEditor-Core.package/MarkdownHeading.class/instance/fontSizes.st
jonaskordt: I thought this too, but I thought it is nice to see that it is a mapping from the level to the size. Hinding the index by using the `OrderedCollection` would hind this mapping because it hides the index. Thus I like it better with the `Dictionary`.

pixunil: Did also thought of semantics, but looking at the line, i only see numbers and it is hard to difference the keys from the values... I find the second option with spaces around the arrows easier to read, but as this is subject-dependant, I'd rather open a poll than discuss long.

```smalltalk
Dictionary newFrom: {1->34. 2->28. 3->22. 4->19. 5->16. 6->14}
Dictionary newFrom: {1 -> 34. 2 -> 28. 3 -> 22. 4 -> 19. 5 -> 16. 6 -> 14}
{34. 28. 22. 19. 16. 14}
```

(okay, that's not really an `OrderedCollection` but an `Array`)

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testMatch1To3Indentations.st
```diff
@@ -0,0 +1,5 @@
+tests - matching
+testMatch1To3Indentations
+
+	#(' #' '  #' '   #') do: [:string |
```
pixunil, line 4: I prefer that we would decide when to use `{}` and when `#()` for array literals.
I think, in cases when we really want an array of symbols, `#()` is appropiate:
https://github.com/hpi-swa-teaching/MarkdownEditor/blob/3448307b24dbe24d713d603e2018d25a12dcf931/packages/MarkdownEditor-Tests.package/MarkdownEditorTest.class/instance/expectedFailures.st#L5
In other cases, like this test, I would prefer `{}`.

Also applies to the other multi-case tests.

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testFindLevelOfHeadingLevels1To6.st
```diff
@@ -0,0 +1,5 @@
+tests
+testFindLevelOfHeadingLevels1To6
+	
+	#('#' 1 '##' 2 '###' 3 '####' 4 '#####' 5 '######' 6) pairsDo: [:string :level |
```
pixunil: Here I would prefer an `Dictionary`, because you map input strings to expected heading levels. Also, the `->` would help to draw the boundaries between the cases.

Also applies to `testTextAttributesOfLevels1To6`

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testMatchHeadingLevels1To6WithText.st
```diff
@@ -0,0 +1,6 @@
+tests - matching
+testMatchHeadingLevels1To6WithText
+	
+	#('# MiniMetro' '## the fastest game' '### on the planet!' '#### cool play style'
```
pixunil, line 4: Oi, I think this will lead to another polarized discussion so here goes nothing:
I find the idea of MiniMetro-flavored test strings funny, but it adds nothing to the understanding of the tests. Instead, they just make the tests wordy.
I would strive for either

- self-explaining test strings like https://github.com/hpi-swa-teaching/MarkdownEditor/blob/3448307b24dbe24d713d603e2018d25a12dcf931/packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/testBlockStructureParagraphWithHeading.st#L6
- or boring, but short test strings like
  ```smalltalk
  {'# H1'. '##  H2'. ... }
  ```

packages/MarkdownEditor-Core.package/MarkdownHeading.class/instance/attributes.st
```diff
@@ -3,7 +3,6 @@ attributes
 	
 	| attributes |
 	attributes := OrderedCollection new.
-	self level odd ifTrue: [attributes add: TextEmphasis underlined].
-	attributes add: (TextFontReference toFont:
-		(TextStyle default copy fontOfSize: self fontSize)).
+	attributes add: (TextFontReference toFont: self font).
+
```
pixunil: Unnecessary newline

pixunil (changes requested):
The tests are much clearer organized when they are grouped at method-level.
But I have some code-style issues which smell like discussion...

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testMatch1To3Indentations.st
felix-gohla: I thought it is kind of a way to represent that these values remain constant. Why shouldn't `#('...')` be appropriate? The only thing that is kinda annoying is the fact that you cannot really see the separation between the strings.

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/testHeadingFollowedByParagraphWithInlineEmphasis.st
```diff
@@ -4,4 +4,8 @@ testHeadingFollowedByParagraphWithInlineEmphasis
 	self
 		assert: '# Title **bold**', Character cr, 'Line *italic*'
 		parsesToRuns: {8. 8. 1. 5. 8}
-		values: {{TextFontReference toFont: (TextStyle default copy fontOfSize: 26). TextEmphasis underlined}. {TextFontReference toFont: (TextStyle default copy fontOfSize: 26). TextEmphasis underlined. TextEmphasis bold.}. {TextFontReference toFont: (TextStyle default copy fontOfSize: 26). TextEmphasis underlined}. {}. {TextEmphasis italic}}
\ No newline at end of file
+		values: {{self headingFontSized: 34}.
```
felix-gohla, line 7: Indentation discussion? Weird: it looks different on GitHub in comparison to Squeak. Maybe one more tab before the lines below to align it?

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testMatchHeadingLevels1To6WithText.st
felix-gohla:
```smalltalk
isParsedStructureOf: 'Paragraph which is interrupted by', Character cr, '# Train arriving'
```
😋

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testTextAttributesOfLevels1To6.st
```diff
@@ -0,0 +1,6 @@
+tests
+testTextAttributesOfLevels1To6
+
+	#('# level1' 34 '## level2' 28 '### level3' 22 '#### level4' 19 '##### level5' 16
```
felix-gohla: Same with dictionary in here.

felix-gohla (changes requested):
Great! Test styling changes would be nice.

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testMatch1To3Indentations.st
jonaskordt: I like the `#()` for constants.

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testMatchHeadingLevels1To6WithText.st
jonaskordt: I don't think we need to be that serious in the tests. Having a bit of fun with them is nice. Thus I like the Minimetro tests.

jonaskordt (approved):
The only thing I'd like to see changes is the `Dictionary`'s in the tests which were pointed out by Felix and Jakob. Other than that it's great.

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testMatch1To3Indentations.st
pixunil: Where would you draw the boundary between `#()` and `{}`?
We have arrays in many of our tests and were always using `{}`. In most cases, `#()` would also work, except when the `#()` would transform class names into symbols. E.g. it would break
https://github.com/hpi-swa-teaching/MarkdownEditor/blob/3448307b24dbe24d713d603e2018d25a12dcf931/packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/testBlockStructureHeadingWithParagraph.st#L5
Also, `#(1->10 2->8)` would transform to something like `#(1 #-> 10 2 #-> 8)`; they don't support dictionaries in the way `{}` does.

So my proposal stands to only use `#()` when the items are actually symbols. In other cases, `{}` just works as expected.
But if you come up with a simple rule which won't always raise discussions whether to write `{}` or `#()`, I'm also happy.

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/testHeadingFollowedByParagraphWithInlineEmphasis.st
pixunil: It looks perfect aligned in my image?

pixunil (approved):
I think we postpone the `#()` vs. `{}`, and funny test strings discussions. But it's nice to see to lose some weight in exchange for a feature :smile_cat:

felix-gohla (approved):

Issue #10, opened 2019-06-07
Tests are inconsistent in argument ordering


pixunil:
The custom asserts like `assert:parsesToRuns:values:`, `assert:hasRuns:values:` in the `BlockParserTest` and `InlineParserTest` should be renamed and reordered in their arguments.

`MarkdownHeadingTest`, category `tests - level` uses `assert: actual equals: expected`

PR #15, opened 2019-06-14
Refactor/textstyler

felix-gohla:
# TextStyler
The architecture got a little more nice, I think. 🤔 
Nevertheless please discuss the changes on refactoring `MarkdownInlineParser` and `MarkdownBlockParser` to its `...Styler` counterparts.
Jonas already updated the structure in LucidChart.

# VB-Regex
I accidentally also pushed a file Patrick edited on my computer. The `.smalltalk.ston` now explicitly tests the `MarkdownEditor-Tests` package and there excludes the `VB-Regex` package tests from being run in smalltalkCI.


packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/parse..st
```diff
@@ -3,9 +3,9 @@ parse: aStringOrText
 
 	^ self 
 		clean;
-		text: aStringOrText asText;
+		textToParse: aStringOrText asText;
 		tokenize;
 		interpretTokens;
 		attributizeText;
-		text
+		textToParse
```
jonaskordt: Here it becomes very weird that the attribute is called `textToParse` because while in line 6 it is still a text that shall be parsed in line 10 the text already has been parsed, which is hidden by the name of the attribute. Sadly we had to rename it because the new super class already has a `text` but if you have a better idea for the name that would be great!

packages/MarkdownEditor-Core.package/MarkdownBlockParser.class/properties.json
```diff
@@ -10,5 +10,5 @@
 	"name" : "MarkdownBlockParser",
```
pixunil: I find it a bit strange that the classes are still named `...Parser`. Wouldn't they be `MarkdownBlockTextStyler`?

packages/MarkdownEditor-Core.package/MarkdownHeading.class/instance/asText.st
```diff
@@ -3,4 +3,4 @@ asText
 
 	| text |
 	text := Text string: self content attributes: self attributes.
-	^ self inlineParser parse: text
\ No newline at end of file
+	^ self styler styledTextFor: text
```
pixunil: Impulse: The `asText` methods are very similiar, and we could revert to the point where `attributes` defaults to `{}` and `asText` be implemented in the abstract class.
This would involve that every `Block` also defines a "region to apply styler" (like `codeContent`). This is only an idea though...

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/parse..st
pixunil: Better, but not there yet: `textToStyle` (as you don't stop after parsing the text)
Also a (maybe bad) idea: you could have accessors named `text` and `text:` which will use another instance variable.

packages/MarkdownEditor-Core.package/MarkdownTextStyler.class/instance/privateStyle..st
```diff
@@ -2,5 +2,5 @@ private
 privateStyle: aText
 
 	| text |
-	text := (MarkdownBlockParser new) parse: aText asString.
+	text := self parse: aText asText.
```
pixunil: I'm confused about the types.
You say `privateStyle:` gets a `Text`, but you call `asText` on it, which should be a no-op.
`parse:` accepts `aStringOrText`, where we always feed it with a `Text`.
Could the `aText asText` here be removed and the type of `parse:` be adjusted to `aText`?

----------

`parse:` acts more like a `styledTextFor:` (which is already taken). So I would either rename `parse:` to `privateStyledTextFor:` or don't implement `privatStyle:` in the abstract class, but in the concrete ones. The only thing that this method does, is to copy over the runs from styled text to input. That shouldn't make much overhead.

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testConvertingToTextUsesInlineParser.st
```diff
@@ -1,12 +1,12 @@
 tests
 testConvertingToTextUsesInlineParser
 
-	| markdownString mockInlineParser heading result |
+	| markdownString mockInlineStyler heading result |
 	markdownString := '# **Stronger** heading with *emphasis*'.
-	mockInlineParser := MarkdownMockInlineParser new.
```
pixunil, line 6: `MarkdowMockInlineParser` should be removed

packages/MarkdownEditor-Core.package/MarkdownBlock.class/instance/styler..st
```diff
@@ -0,0 +1,4 @@
+accessing
+styler: aMarkdownTextStyler
```
pixunil:
```suggestion
styler: aSHTextStyler
```
We also put a `SHTextStylerST80` in this variable

packages/MarkdownEditor-Core.package/MarkdownBlockParser.class/instance/clean.st
```diff
@@ -0,0 +1,4 @@
+initialize-release
+clean
+
+	self blocks: OrderedCollection new.
```
pixunil:
```suggestion
	self blocks: OrderedCollection new
```

packages/MarkdownEditor-Tests.package/MarkdownInlineParserTest.class/instance/testIsReusable.st
```diff
@@ -0,0 +1,9 @@
+tests - cleaning
+testIsReusable
+
+	| styledResult |
+	styledResult := inlineParser styledTextFor: '**HPI MiniMetro**' asText.
+	self assert: styledResult  hasRuns: {17} values: {{TextEmphasis bold}}.
```
pixunil:
```suggestion
	self assert: styledResult hasRuns: {17} values: {{TextEmphasis bold}}.
```
Double spaces (also the assert below)

pixunil (changes requested):
Hm, some parts are getting easier, but the transition is not quite complete imo. It would be nice to see that they are no `Parser` (so other name and no `parse:`), but really `TextStyler`'s.

-----

> I accidentally also pushed a file Patrick edited on my computer. The .smalltalk.ston now explicitly tests the MarkdownEditor-Tests package and there excludes the VB-Regex package tests from being run in smalltalkCI.

Your description confuses me - the `VB-Regex` thingy is in #12 too, right?

packages/MarkdownEditor-Core.package/MarkdownBlockParser.class/properties.json
jonaskordt: Yes. We thought so too but we wanted to discuss this with the group as we weren't sure if everyone was onboard with that.

packages/MarkdownEditor-Core.package/MarkdownInlineParser.class/instance/parse..st
jonaskordt: Well but `textToStyle` doesn't make sense either because in line 10 it is already styled. Thus at that point it is no longer a text to be styled.

packages/MarkdownEditor-Core.package/MarkdownTextStyler.class/instance/privateStyle..st
jonaskordt: We were confused about the types too  :D 

The `asText` in line 5 is not necessary. That's correct. 

The main reason why we kept the `aStringOrText` in the `parse:` methods is because we always test them with strings and didn't want to deal with changing all the tests. I personally think it is a good idea to change it to always use Texts and never use strings especially since we are changing the whole classes to TextStylers which don't care about strings.

I think this PR needs some more time in our next iteration since the requested changes are too big for this one.

packages/MarkdownEditor-Core.package/MarkdownCodeBlock.class/instance/asText.st
```diff
@@ -2,7 +2,7 @@ converting
 asText
 
 	| code result |
-	code := self smalltalkStyler styledTextFor: self codeContent asText.
+	code := self styler styledTextFor: self codeContent asText.
```
jonaskordt: Here we might get some problems since this method was changed in #12

jonaskordt (changes requested):
Die Coverage ist noch rot weil wir vergessen haben `MarkdownInlineTextStyler>>attributizeText` zu entfernen. Dafür jetzt einen extra Commit zu machen muss ja nicht sein, fall es noch weitere Changes gibt. Also wird das geändert wenn es sowieso Änderungen gibt oder kriegt seinen eigenen Commit wenn es keine weiteren Änderungen mehr gibt.

packages/MarkdownEditor-Tests.package/MarkdownInlineTextStylerTest.class/instance/assertRuns.values.areStyleOf..st
```diff
@@ -1,6 +1,7 @@
 accessing
-assert: aText hasRuns: runs values: values
+assertRuns: runs values: values areStyleOf: aText
```
pixunil, line 2: This is a suggestion which resolves #10. I'm not happy with the naming, but this is the best thing I could have thought of...
When we're happy, we should also use this in the block tests

packages/MarkdownEditor-Tests.package/MarkdownBlockParserTest.class/instance/testProcessLineCreatesParagraph.st
```diff
@@ -1,7 +0,0 @@
-tests
-testProcessLineCreatesParagraph
```
pixunil, line 2: We removed 3 tests concerning the old `processLine:` because they effectively test the block structure on a more granular level. We also found it strange that they only use headings and paragraphs.

packages/MarkdownEditor-Tests.package/MarkdownInlineParserTest.class/instance/testAttributizeText.st
```diff
@@ -1,10 +0,0 @@
-tests - attributes
-testAttributizeText
```
pixunil, line 2: We removed a test about `attributizeText` as this should be covered by testing `privateStyle`.

pixunil (commented):
I just add some comments to clarify a bit our thoughts.
We now try to feed the parsing component with the string. It will be chopped into chunks (lines or tokens). They are used to create an intermediate representation.
The representation are either blocks or attributes. With these, we style the text.

Also, I find it a bit of an anti-pattern to use a method object here, as a method object should be used only once, but a text styler is used multiple times. Thus, we strive to avoid putting information into instance variables. This works for the block styler, but the inline styler is hopeless in its current implementation.

packages/MarkdownEditor-Tests.package/MarkdownInlineTextStylerTest.class/instance/assertRuns.values.areStyleOf..st
felix-gohla: `assertRuns: runs values: values areAttributesOf: aText`? Maybe not this great, too. It lacks the intention that the text is styled by the inlineParser (which should be named inlineStyler in here as well). Why don't you style outside of this method?
Funny: GitHub displays this as MarkdownInlinePARSERTest...

packages/MarkdownEditor-Tests.package/MarkdownInlineTextStylerTest.class/instance/testInterpretTokensWithoutEmphasis.st
```diff
@@ -0,0 +1,6 @@
+tests - interpretation
+testInterpretTokensWithoutEmphasis
+
+	inlineParser parse: 'foo'.
```
felix-gohla: Shouldn't we only test interpretation of the tokens in here? It conflicts some way with `testParseNormal `.

packages/MarkdownEditor-Core.package/MarkdownBlockTextStyler.class/instance/process.into..st
```diff
@@ -0,0 +1,8 @@
+parsing
+process: aString into: aCollectionOfMarkdownBlocks
```
felix-gohla, line 2: I like the decomposition of the `MarkdownBlockTextStyler`'s parse method. 👍

packages/MarkdownEditor-Core.package/MarkdownInlineTextStyler.class/instance/attributizeText.st
```diff
@@ -3,7 +3,7 @@ attributizeText
 
 	self attributes do: [:attribute |
 		attribute textAttributes do: [:textAttribute |
-			self text 
+			self textToParse
```
felix-gohla: Yep, please remove!

felix-gohla (commented):
Looks pretty good to me.

Should `privateStyle:` be `subclassResponsibility` in MarkdownTextStyler?

packages/MarkdownEditor-Tests.package/MarkdownInlineTextStylerTest.class/instance/testInterpretTokensWithoutEmphasis.st
jonaskordt: I think the test content is fine but the naming and categories could be improved. The `parse` method does less than before (it stops after interpreting the tokens) and thus what we used to test by first tokenizing and then analyzing those tokens we now test by simply calling parse. We could instead also call `tokenize:` and then `interpretTokens` and then compare with `inlineParser attributes`. Maybe that would show better what we are actually testing.

`testParseNormal` should really be renamed to `testStyleNormal` as we test the `privateStyle` method and no longer `parse`.

packages/MarkdownEditor-Tests.package/MarkdownInlineTextStylerTest.class/instance/assertRuns.values.areStyleOf..st
jonaskordt: Well the intention of this method is to encapsulate everything that is the same for all the tests in which it is used. `privateStyle` always has to be called thus I think it makes sense to have it in the method. Otherwise it would have to be in all tests which would make it more visible but also bulk up the tests. But if we decide that it is more important to see what happens then that would be a good option too.

packages/MarkdownEditor-Tests.package/MarkdownInlineTextStylerTest.class/instance/testInterpretTokensWithoutEmphasis.st
felix-gohla: Please rename the test's instance variables to `inlineTextStyler` or `inlineStyler`.

packages/MarkdownEditor-Core.package/MarkdownBlockTextStyler.class/instance/buildTextFrom..st
```diff
@@ -0,0 +1,6 @@
+parsing
+buildTextFrom: aCollectionOfMarkdownBlocks
```
felix-gohla: Here we could just return the `runs` of the text being created. Furthermore, renaming this method to something like `attributesFor: ...` would be nice.

packages/MarkdownEditor-Tests.package/MarkdownInlineTextStylerTest.class/instance/assertRuns.values.areStyleOf..st
jonaskordt: `privateStyle:` is now always called outside of the method. This makes it very visible what is happening.

packages/MarkdownEditor-Tests.package/MarkdownBlockTextStylerTest.class/instance/setUp.st
```diff
@@ -0,0 +1,4 @@
+running
+setUp
+
+	blockParser := MarkdownBlockTextStyler new
```
kiragrammel: Why do we still have a blockParser here? I thought they should all be replaced by blockTextStyler.

packages/MarkdownEditor-Core.package/MarkdownBlockTextStyler.class/instance/process.into..st
jonaskordt: Thanks!

kiragrammel (approved):

packages/MarkdownEditor-Tests.package/MarkdownBlockTextStylerTest.class/instance/setUp.st
jonaskordt: Haha you literally approved and commented seconds after I pushed the changes. This is changed there.

jonaskordt (approved):

packages/MarkdownEditor-Core.package/MarkdownBlockTextStyler.class/instance/privateStyle..st
```diff
@@ -0,0 +1,6 @@
+parsing
```
pixunil, line 1: Hmm, the category seams to be wrong. `SHTextStyler` puts `privateStyle:` into `private`, whereas I would put `privateStyle:` and `buildAttributeRunsFrom:` into the same category, for example `styling`. What category do you think is best?

Same problem in the `InlineTextStyler`

packages/MarkdownEditor-Core.package/MarkdownBlockTextStyler.class/instance/buildAttributeRunsFrom..st
```diff
@@ -0,0 +1,6 @@
+parsing
+buildAttributeRunsFrom: aCollectionOfMarkdownBlocks
+	
+	^ (aCollectionOfMarkdownBlocks
```
pixunil, line 4: I'd prefer to put the larger expression into a variable to avoid overlooking `runs`.

```smalltalk
| text |
text := (aCollectionOfMarkdownBlocks
	inject: Text new
	into: [:text :block | text append: block asText]).
^ text runs

**Review by pixunil — APPROVED** [review]
The categories could be better divided into `parsing` and `styling` in both stylers.
I even would suggest to remove the base class `MarkdownTextStyler` as it adds little benefit (`reset` and `privateFormat:`)
Doing a soft approve (I'll wait for possible request from @lpfennigschmidt and then would recategorize / implement suggestions)

File: packages/MarkdownEditor-Tests.package/MarkdownInlineTextStylerTest.class/instance/assertRuns.values.areStyleOf..st (hunk 10 above)
Comment (pixunil) [reply packages/MarkdownEditor-Tests.package/MarkdownInlineTextStylerTest.class/instance/assertRuns.values.areStyleOf..st:2]: Applying `privateStyle:` outside of the method should work for the inline text styler, but for the block text styler we would either remove the assertion that the string remains unmodified or have a better name which implies that the styler is used...

File: packages/MarkdownEditor-Tests.package/MarkdownBlockTextStylerTest.class/instance/assert.parsesToRuns.values..st [hunk 20]
```diff
@@ -0,0 +1,10 @@
+accessing
+assert: markdownString parsesToRuns: runs values: values
+
+	| text |
+	text := markdownString asText.
```
felix-gohla, line 5: This is still the incorrect parameter order for the assert statement. What about `assertRuns: runs values: values areStyledResultOf:`? This should imply that the string will be styled...

packages/MarkdownEditor-Core.package/MarkdownBlockTextStyler.class/instance/privateStyle..st
felix-gohla: `styling` would be fine with me.

felix-gohla (commented):
In my opinion, this can be merged when the parameter order for the `MarkdownBlockStyler` tests are fixed.

packages/MarkdownEditor-Tests.package/MarkdownBlockTextStylerTest.class/instance/assert.parsesToRuns.values..st
pixunil: Correct. We have in the heading tests also one confused order. (#10) But I would open a second PR for it, because this one gets to large to put more stuff into it...

PR #21, opened 2019-06-23
Refines code style

pixunil:
- [x] Uses` assertRuns:values:areStyledResultOf:` in `BlockTextStylerTest` instead of the old `assert:parsesToRuns:values:`
- [x] Removes assert-sugars in `HeadingsTest`
- [x] Replaces `#()` with `{}`
- [x] Removes some double spaces
- [x] Categorizes styling methods
- [x] Decide code style for dictionaries in tests
- [x] Decide whether to remove `MarkdownTextStyler`
- [x] Add inline text styler tests (apply theme + styling / formatting)

(Sorry for the forced pushs, but I did some stupid things that don't deserve their own fix-commit)


packages/MarkdownEditor-Tests.package/MarkdownEmphasisDelimiterTest.class/instance/testLength.st
```diff
@@ -1,13 +1,6 @@
...
-	self assert: 1 equals: delimiter length.
-	
-	delimiter := MarkdownEmphasisDelimiter newFor: (self readStreamOn: '****').
-	self assert: 4 equals: delimiter length.
-	
-	delimiter := MarkdownEmphasisDelimiter newFor: (self readStreamOn: '*****').
-	self assert: 5 equals: delimiter length
-	
\ No newline at end of file
+	{'*'->1. '**'->2. '***'->3. '****'->4. '*****'->5} do: [:case | | delimiter |
```
pixunil: This is an alternative style I'm proposing here. If you prefer `Dictionary newFrom: {}`, I'll change it here, otherwise I would replace the old occurrences of `Dictionary newFrom:`.

------

There is also another method which tests the length, but with `calculateLength:`. I don't understand why they are necessary and test funny things like the length of a right-flanking delimiter...

https://github.com/hpi-swa-teaching/MarkdownEditor/blob/6d24bdf8863a394d9acbf6998b3ad181759eb5aa/packages/MarkdownEditor-Tests.package/MarkdownEmphasisDelimiterTest.class/instance/testLengthCalculation.st#L1-L8

packages/MarkdownEditor-Core.package/MarkdownBlockTextStyler.class/instance/buildAttributeRunsFrom..st
```diff
@@ -1,6 +1,8 @@
...
 buildAttributeRunsFrom: aCollectionOfMarkdownBlocks
-	
-	^ (aCollectionOfMarkdownBlocks 
-		inject: Text new 
-		into: [:text :block | text append: block asText]) runs
+
+	| text |
+	text := Text new.
+	aCollectionOfMarkdownBlocks do: [:block |
+		text append: block asText].
```
pixunil: I liked `inject:into:` but I think a simple `do:` shows that there is no magic happening here...

pixunil (commented):
Here some semi-random code style changes.
The proposal to remove `MarkdownTextStyler` and move `reset` and `privateFormat:` into the subclasses still stands.

packages/MarkdownEditor-Tests.package/MarkdownEmphasisDelimiterTest.class/instance/testLength.st
jonaskordt: I really like that fact that using `keysAndValuesDo: [:string :length | ...]` gives the `string` and the `level` a name. Having `case key` and `case value` is really nondescriptive. 
But that doesn't work with this kind of Dictionary notation. I'd prefer to keep it like it is with the other Dictionaries.

jonaskordt: I don't see the problem with the `testLengthCalculation`. It tests multiple lengths to make sure that not only one length is supported (this is a perfect example of how TDD works since first it returned 1, then it supported 2 and then it supported everything else, then it supports text after the delimiter) and it also checks if it works for a closing delimiter since that is a litte different.

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testMatchHeadingLevels1To6WithText.st
```diff
@@ -1,6 +1,5 @@
 tests - matching
 testMatchHeadingLevels1To6WithText
-	
-	#('# MiniMetro' '## the fastest game' '### on the planet!' '#### cool play style' 
-	'##### endless fun' '###### universal gratitude of passengers') do: [:string |
+
+	{'# H1'. '## H2'. '### H3'. '#### H4'. '##### H5'. '###### H6'} do: [:string |
```
jonaskordt, line 4: Let us have a little fun with test strings... This one really didn't mess with understandability...

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testMismatchHeadingLevels1To6WithUnspacedText.st
```diff
@@ -1,6 +1,5 @@
 tests - matching
 testMismatchHeadingLevels1To6WithUnspacedText
 
-	#('#MiniMetro' '##the fastest game' '###on the planet!' '####cool play style'
-	'#####endless fun' '######universal gratitude of passengers') do: [:string | 
+	{'#H1'. '##H2'. '###H3'. '####H4'. '#####H5'. '######H6'} do: [:string |
```
jonaskordt: same here...

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testTextAttributesOfLevels1To6.st
```diff
@@ -1,7 +1,8 @@
 tests
 testTextAttributesOfLevels1To6
 	
-	(Dictionary newFrom: {'# level1'->34. '## level2'->28. '### level3'->22.
-	'#### level4'->19. '##### level5'-> 16. '###### level6'-> 14}) 
-		keysAndValuesDo: [:string :size |
-			self assert: string asTextHasAttributes: {self headingFontSized: size}]
+	(Dictionary newFrom: {'# H1'->34. '## H2'->28. '### H3'->22.
+	'#### H4'->19. '##### H5'->16. '###### H6'->14}) 
+		keysAndValuesDo: [:string :size | | parsedText |
```
jonaskordt, line 6: This is how I think `keysAndValuesDo:` improves the test!

jonaskordt (changes requested):
As I told you face to face I am all for the `{...}` now so that's great.

I don't like the Dictionary notation you suggested. See the comments for more details on that.

Also I don't want to remove the `MarkdownTextStyler`. No need to introduce the reset method to both of them.

lpfennigschmidt (changes requested):
I'm with Jonas on the Dictionaries- and keysAndValuesDo-Discussion.
I also get an error saying that MarkdownTextStyler should implement privateStyle. I guess this is just because I have a lot of (110 or so) old instances flying around, but maybe that's also something we should look into. But this also lets me side with Jakob on just removing the whole class.

jonaskordt:
> I'm with Jonas on the Dictionaries- and keysAndValuesDo-Discussion.
> I also get an error saying that MarkdownTextStyler should implement privateStyle. I guess this is just because I have a lot of (110 or so) old instances flying around, but maybe that's also something we should look into. But this also lets me side with Jakob on just removing the whole class.

jonaskordt:
I don't think we should remove a class only because you get errors with old instances. Those errors won't go away when the class is gone since then it will still look for the implementation but not finde any at all.

I think the `MarkdownTextStyler` should have `privateStyle` as `subclassResponsibility` as that one is most definitely required. 

About the old instances: Why are you getting those? I don't get them so it's hard for me to see why they would appear.

packages/MarkdownEditor-Tests.package/MarkdownEmphasisDelimiterTest.class/instance/testLength.st
pixunil: Okay, after digging the code a bit more, I saw that `testLengthCalculation` only sets the `length` indirectly, so it's okay to have tests for both. But maybe this can become simpler after revisiting the `InlineTextStyler`...

packages/MarkdownEditor-Core.package/MarkdownBlockTextStyler.class/instance/buildAttributeRunsFrom..st
felix-gohla: I personally don't see a improvement in this.

packages/MarkdownEditor-Core.package/MarkdownCodeBlock.class/instance/asText.st
```diff
@@ -3,7 +3,7 @@ asText
 
 	| code result |
 	code := self codeContent asText.
-	self styler privateStyle: code.
+	self styler privateStyle: code. "We use the private method here as the public interface of SHTextStyler does not have a method which only styles and gives the value back to the caller"
```
felix-gohla: Please put the comment above the line.

packages/MarkdownEditor-Tests.package/MarkdownEmphasisDelimiterTest.class/instance/testLength.st
```diff
@@ -1,13 +1,7 @@
...
-	self assert: 1 equals: delimiter length.
-	
-	delimiter := MarkdownEmphasisDelimiter newFor: (self readStreamOn: '****').
-	self assert: 4 equals: delimiter length.
-	
-	delimiter := MarkdownEmphasisDelimiter newFor: (self readStreamOn: '*****').
-	self assert: 5 equals: delimiter length
-	
\ No newline at end of file
+	(Dictionary newFrom: {'*'->1. '**'->2. '***'->3. '****'->4. '*****'->5})
```
felix-gohla, line 4: 👍

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testMismatchHeadingLevels1To6WithUnspacedText.st
```diff
@@ -1,6 +1,6 @@
 tests - matching
 testMismatchHeadingLevels1To6WithUnspacedText
 
-	#('#MiniMetro' '##the fastest game' '###on the planet!' '####cool play style'
-	'#####endless fun' '######universal gratitude of passengers') do: [:string | 
+	{'#hashtag'. '##missing space'. '###no heading'. '####just text'. '#####five hashes'.
```
felix-gohla: When you really want to delete the funny test strings then please be consistent. I know, the first ones are self explaining, but not the last ones.

felix-gohla (changes requested):

felix-gohla:
> I'm with Jonas on the Dictionaries- and keysAndValuesDo-Discussion.
> I also get an error saying that MarkdownTextStyler should implement privateStyle. I guess this is just because I have a lot of (110 or so) old instances flying around, but maybe that's also something we should look into. But this also lets me side with Jakob on just removing the whole class.

I don't think we should remove the class. It would introduce some code duplication. But I'll investigate the fact that old instances aren't trashed by the GC.

felix-gohla:
> > I'm with Jonas on the Dictionaries- and keysAndValuesDo-Discussion.
> > I also get an error saying that MarkdownTextStyler should implement privateStyle. I guess this is just because I have a lot of (110 or so) old instances flying around, but maybe that's also something we should look into. But this also lets me side with Jakob on just removing the whole class.
> 
> I don't think we should remove the class. It would introduce some code duplication. But I'll investigate the fact that old instances aren't trashed by the GC.

Ok, I found the reason. AutoTDD seems to keep references to the tested objects. Therefore, closing AutoTDD and running the GC should do the job.

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testMismatchHeadingLevels1To6WithUnspacedText.st
pixunil: Okay, the last one `######long fenced railway gate` was meant as a compromise - as the reader should understand what's happening after the first five cases and then I don't care about the content.
Nevertheless, it was wrong before that the spaced and unspaced version use the same strings

packages/MarkdownEditor-Core.package/MarkdownBlockTextStyler.class/instance/buildAttributeRunsFrom..st
pixunil: So, is it a request to change it back or are you neutral which version we should take?

jonaskordt (approved):
I'm still not happy with the test strings being removed. We can expect whoever looks at the tests to know Markdown. I think it's okay to have a little fun with that.

I won't block this from being merged because of that but I'd really like to see it changed back.

packages/MarkdownEditor-Core.package/MarkdownBlockTextStyler.class/instance/buildAttributeRunsFrom..st
felix-gohla: Yes it is. I think the `inject` statement fits the solution a little more.

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testMismatchHeadingLevels1To6WithUnspacedText.st
felix-gohla: Yeah, that's an improvement, but the last element doesn't add anything to the test's semantic...

packages/MarkdownEditor-Tests.package/MarkdownHeadingsTest.class/instance/testMatchHeadingLevels1To6WithText.st
kiragrammel: I like it like that. Especially that it fits in one line.

kiragrammel (approved):

lpfennigschmidt (approved):

packages/MarkdownEditor-Core.package/MarkdownBlockTextStyler.class/instance/buildAttributeRunsFrom..st
pixunil: I tried to make a version which uses `inject:into:` and shows explicitly that we take the runs of a text, but it always resulted in overhead.

```smalltalk
| text |
text := aCollectionOfMarkdownBlocks inject: Text new into: [:accumulatedText :block | accumulatedText append: block asText].
^ text runs
```

I'm also wary because we don't return a new object by using `inject:into:`, but modify one text object. In the normal cases, you always make a new object (like in `inject: 0 into: [:acc :number | acc + number]`), but I dunno whether you would do it in our case...

jonaskordt: I don't see the problem with the `inject:into:` code in your last comment. Obviously it needs line breaks but there is no overhead that I can see.
I think it's okay to modify instead of creating a new object all the time.

felix-gohla (approved):
Noice. 😊

jonaskordt (changes requested):
Even though our editor doesn't use the `InlineTextStyler` as a TextStyler by itself we still make it available as a TextStyler which can be used. Thats why we had the PR to make it usable by itself. But that includes that at least the existence of the `reset` and `privateFormat:` methods need to be tested.

jonaskordt (approved):
Approval vorbehaltlich der Korrektur des einen Typos im Testnahmen den Lara genannt hat.

Issue #22, opened 2019-06-25
InlineStyler doesn't work properly when used as a TextStyler by itself.


jonaskordt:
The `InlineStyler` currently is not allowed to remove attributes from the text it is given. When it is used as a TextStyler just by itself that is bad.

Example: `**bold*` -> \*_\*bold\*_
This is still correct. But now the Text has the italic attribute. 
Now we type another `*`. 
Then we would expect **\*\*bold\*\*** but since we still have the italic attribute and don't remove it we get **\*_\*bold\*_\***. 

The reason why the InlineStyler is currently not allowed to remove attributes is 
```smalltalk
MarkdownHeading>>asText

	| text |
	text := Text string: self content attributes: self attributes.
	self styler privateStyle: text.
	^ text
```
Here the heading attributes are added first and then the inline attributes are added. By changing the order (adding inline attributes first and then adding heading attributes onto of that) we could change it so that the InlineTextStyler is allowed to remove attributes. This would be cool because then it could be used as a TextStyler by itself.

PR #23, opened 2019-06-25
Makes InlineTextStyler usable by itself

jonaskordt:
-addresses #22
-switches order of adding attributes in `MarkdownHeading>>asText`
-`MarkdownInlineTextStyler` now removes any attributes from the text it is given

I think it is important that the `MarkdownInlineTextStyler` is usable by itself, even though we don't use it like that, because otherwise it makes no sense for it to be a TextStyler.


packages/MarkdownEditor-Core.package/MarkdownInlineTextStyler.class/instance/privateStyle..st
```diff
@@ -2,5 +2,6 @@ parsing
 privateStyle: aText
 	
 	| attributes |
+	aText runs: (RunArray runs: {aText size} values: {{}}).
```
jonaskordt: This line is the one which removes the previous attributes. I couldn't find a cleaner way to do this. If you have any suggestions please do tell.

Sadly we don't have a `Text` where we can get `runs` from to simply overwrite the `runs` of `aText` like we can in `MarkdownBlockTextStyler>>privateStyle`.

felix-gohla: How about:
`aText addAttribute: TextEmphasis normal.`?

felix-gohla:
I don't really get the point of this PR... Why should I use it by itself? I read #22. The wording for your PR is confusing. Maybe we could rename it to "Allow InlineTextParser to remove attributes".

jonaskordt:
While you might not have an application for using it by itself right now someone in the future might. There are often occurrences where only inline text should be parsed as allowing Headings and stuff would be too attention grabbing. Since we are creating a `MarkdownInlineTextStyler` we want it to be possible to be used that way. The title of the PR makes perfect sense.

packages/MarkdownEditor-Core.package/MarkdownInlineTextStyler.class/instance/privateStyle..st
jonaskordt: Thanks

packages/MarkdownEditor-Tests.package/MarkdownInlineTextStylerTest.class/instance/testPreviousAttributesGetRemoved.st
```diff
@@ -1,10 +1,10 @@
 tests - parsing
-testPlainStylingKeepsAttributes
+testPreviousAttributesGetRemoved
```
pixunil, line 2: The `BlockTextStyler` should also remove attributes. Could you add a test for it?

pixunil (approved):

kiragrammel (approved):

PR #24, opened 2019-06-25
HTML Comment Blocks

jonaskordt:
This is ready and can be reviewed even though the tests in 5.0 are failing as we plan to no longer support 5.0. No reason why you couldn't review it yet though.

Obviously merging will have to wait for the removal of 5.0 support.


packages/MarkdownEditor-Tests.package/MarkdownCommentBlockTests.class/instance/testMismatchGTAfterOpeningTag.st
```diff
@@ -0,0 +1,4 @@
+tests - matching
+testMismatchGTAfterOpeningTag
```
felix-gohla: Please no abbreviations in test names.

packages/MarkdownEditor-Tests.package/MarkdownCommentBlockTests.class/instance/testMismatchMinusGTAfterOpeningTag.st
```diff
@@ -0,0 +1,4 @@
+tests - matching
+testMismatchMinusGTAfterOpeningTag
```
felix-gohla: Same here.

felix-gohla (commented):
Perfect. Works great.

lpfennigschmidt (approved):
:D

packages/MarkdownEditor-Core.package/MarkdownCommentBlock.class/class/matches..st
```diff
@@ -0,0 +1,7 @@
+testing
+matches: aString
+	
+	^ (aString beginsWith: '    ') not
+		and: [aString withBlanksTrimmed beginsWith: '<!--']
+		and: [((aString includesSubstring: '-->') not)
+			or: [(aString withBlanksTrimmed findString: '-->') >= 5]]
```
kiragrammel: Can we avoid the magic number here? 🤔 Or do you think it is for everyone clear?

packages/MarkdownEditor-Core.package/MarkdownCommentBlock.class/instance/commentEnd.st
```diff
@@ -0,0 +1,6 @@
+accessing
+commentEnd
+
+	^ self isClosed
+		ifTrue: [(self content findString: '-->') + 2]
```
kiragrammel: The same with the 2.

kiragrammel (approved):
Some class comments are missing e.g. for all MarkdownBlock classes. I don't know if we need class comments for the test classes. 🤔

packages/MarkdownEditor-Core.package/MarkdownCommentBlock.class/class/matches..st
lpfennigschmidt: We could say '<!--' size +1, as that is where the number comes from. The starting position of the closing tag may not be at the first 4 characters, if it is in the same line as the opening tag. Would that be better?

packages/MarkdownEditor-Core.package/MarkdownCommentBlock.class/instance/commentEnd.st
lpfennigschmidt: Here it could be '-->' size -1. I don't know if all this index-shifting makes it more clearer

lpfennigschmidt (approved):
And yes, we should write class comments ^^

packages/MarkdownEditor-Core.package/MarkdownCommentBlock.class/class/matches..st
jonaskordt: This makes me want to introduce 
```
MarkdownCommentBlock>>openingTag

    ^ '<!--'
```
on the class side and the same for `MarkdownCommentBlock>>closingTag`. Then we could use that and ask for the size and then the index shifting would make sense. We could also use those in `matches:` and `isClosed`. That way we could get rid of the magic numbers and magic strings at the same time.
Here we could even change it to `(aString withBlanksTrimmed findString: '-->') > self openingTag size` and get rid of the index shifting. In `MarkdownCommentBlock>>commentEnd` the index shifting would still be necessary sadly.

Obviously this is not quite simple design anymore and the argument of changing magic strings doesn't really apply because we don't really expect the Markdown and/or HTML spec to change but on the software engineering side of things it would be an improvement I think.
This would also make it into a generic method for a lot of HTML blocks.

jonaskordt: Line 6 and 7 are actually an implication. So we could use `==>` instead of the combination of `not` and `or:`. What do you think?

packages/MarkdownEditor-Core.package/MarkdownCommentBlock.class/instance/commentStart.st
```diff
@@ -1,4 +1,4 @@
 accessing
 commentStart
 
-	^ self firstLine findString: '<!--'
\ No newline at end of file
+	^ self firstLine findString: self class openingTag
```
jonaskordt: We could consider adding an accessor for the opening and closing tag on the object side too which would just call the class accessor. That way calling it on the object side would get a little cleaner but I'm not sure if that is necessary.

packages/MarkdownEditor-Core.package/MarkdownCommentBlock.class/instance/asText.st
```diff
@@ -0,0 +1,10 @@
+converting
+asText
+
+	| text |
+	text := self content asText.
+	text 
+		addAttribute: (TextColor color: Color lightGray)
```
pixunil: We could make this similiar to `MarkdownHeading`.
Move `TextColor` into a constant `textAttribute`

packages/MarkdownEditor-Core.package/MarkdownParagraph.class/class/matches..st
```diff
@@ -0,0 +1,4 @@
+matching
+matches: aString
```
pixunil: Either we make `MarkdownBlock>>matches` a `subclassResponsibility` or we modify the caller to:
```smalltalk
determineMatchingBlockClass: aString

	^ {MarkdownHeading. MarkdownCodeBlock. MarkdownCommentBlock}
		detect: [:blockClass | blockClass matches: aString]
		ifNone: [MarkdownParagraph]
```
I'd prefer the later one, as it encodes explicitly the "catch-all" behaviour of `MarkdownParagraph`.

Also, Jonas and I discussed whether we rename them to `mayBeginWith:`, too

packages/MarkdownEditor-Tests.package/MarkdownBlockTextStylerTest.class/instance/testBlockStructureCommentBlockNotInterruptedByHeading.st
```diff
@@ -0,0 +1,6 @@
+tests - blocks
+testBlockStructureCommentBlockNotInterruptedByHeading
+
+	self
+		assert: {MarkdownCommentBlock}
+		isParsedStructureOf: '<!--', Character cr, '# Not a heading', Character cr, '-->'
```
pixunil, line 6: I miss a test that comments will consume code blocks and a test that code blocks will consume comments. I can't deduce a fitting theory that we are covering it somewhere else (but if you have a nice explanation, I'll take that)

packages/MarkdownEditor-Tests.package/MarkdownCommentBlockTests.class/instance/assertRuns.values.areStyleOf..st
```diff
@@ -0,0 +1,6 @@
+accessing
+assertRuns: runs values: values areStyleOf: aText
```
pixunil, line 2: We really should implement this helper somewhere more central... two times the same code is okay, but a third time is a pattern

packages/MarkdownEditor-Tests.package/MarkdownCommentBlockTests.class/instance/testClosingTagInFirstLineCloses.st
```diff
@@ -0,0 +1,6 @@
+tests - extendability
+testClosingTagInFirstLineCloses
+
+	| commentBlock |
+	commentBlock := MarkdownCommentBlock newFrom: '<!-- MiniMetro -->', Character cr.
```
pixunil: 'This is the granularity where I prefer the self-explaining version. I honestly laughed about the block structure tests, but here it feels off.

```suggestion
	commentBlock := MarkdownCommentBlock newFrom: '<!-- Comment closed at the same line -->', Character cr.

File: packages/MarkdownEditor-Tests.package/MarkdownCommentBlockTests.class/instance/testClosingTagFollowedByTextCloses.st [hunk 11]
```diff
@@ -0,0 +1,7 @@
+tests - extendability
+testClosingTagFollowedByTextCloses
+
+	| commentBlock |
+	commentBlock := MarkdownCommentBlock newFrom: '<!-- MiniMetro', Character cr.
+	commentBlock append: 'is great --> more Text'.
```
pixunil, line 6:
```smalltalk
	commentBlock := MarkdownCommentBlock newFrom: '<!-- Comment with', Character cr.
	commentBlock append: 'multiple lines --> Text after end tag'.
```
Applies to other `isClosed` and `commentEnd` tests

packages/MarkdownEditor-Tests.package/MarkdownCommentBlockTests.class/instance/testCommentStart0To3Indentation.st
```diff
@@ -0,0 +1,6 @@
+tests - comment
+testCommentStart0To3Indentation
+
+	(Dictionary newFrom: {'<!--'->1.' <!--'->2. '  <!--'->3. '   <!--'->4}) keysAndValuesDo: [:commentLine :start |
+		self assert: start equals: (MarkdownCommentBlock newFrom: commentLine) commentStart]
```
pixunil: I'd prefer to put the comment block into an own variable to avoid overlooking `commentStart` and making the line shorter.

Also, please put `keysAndValuesDo:` on a new line

packages/MarkdownEditor-Tests.package/MarkdownCommentBlockTests.class/instance/testMismatchOverlappingOpeningAndClosingTag.st
```diff
@@ -0,0 +1,5 @@
+tests - matching
+testMismatchOverlappingOpeningAndClosingTag
+
+	self deny: (MarkdownCommentBlock matches: '<!-->').
```
pixunil: after reading the spec, I think that the rules
1. Up to 3 indentation
2. Begins with `<!--`

are sufficient. It differs for inline comments though, where `<!-->` and `<!--->` and `<!-- -- -->` are forbidden.
So I think these tests are wrong and you can remove some lines in `matches:`. But maybe I just misread the spec...

pixunil (changes requested):
The code reads like a charm, the only puzzling moment is - like always - `matches`. I hope that my interpretation of the spec is correct, and that you can simplify it.

packages/MarkdownEditor-Core.package/MarkdownCommentBlock.class/class/matches..st
pixunil: I would say that `openingTag` is in the spirit of simple design: Make it clear what you are doing, by expressing it with the least possible _complexity_. (And number of methods is only a hint to complexity, not a hard criterium)

packages/MarkdownEditor-Core.package/MarkdownParagraph.class/class/matches..st
jonaskordt: I changed the code. I do think renaming it to `mayBeginWith:` would be good but I don't want to change it without feedback from others.

packages/MarkdownEditor-Tests.package/MarkdownCommentBlockTests.class/instance/assertRuns.values.areStyleOf..st
jonaskordt: I already suggested a `TestCase` subclass which contains all the helper functions that we frequently use. Let's discuss this as a group.

packages/MarkdownEditor-Tests.package/MarkdownCommentBlockTests.class/instance/testClosingTagFollowedByTextCloses.st
jonaskordt: Meh. I think it's pretty clear like it is...

packages/MarkdownEditor-Tests.package/MarkdownCommentBlockTests.class/instance/testMismatchOverlappingOpeningAndClosingTag.st
jonaskordt: Hm seems like you are right. I think it's confusing that they have different rules abut trying it with https://spec.commonmark.org/dingus/ gives the result you described.

(I changed it already but didn't resolve it to avoid confusion)

packages/MarkdownEditor-Tests.package/MarkdownCommentBlockTests.class/instance/assertRuns.values.areStyleOf..st
jonaskordt: I think this is something for a different PR. I created #28 so that we don't forget.

packages/MarkdownEditor-Core.package/MarkdownCommentBlock.class/class/closingTag.st
```diff
@@ -0,0 +1,4 @@
+accessing
```
pixunil:
```suggestion
constants
```
Applies also to `openingTag` and the variants at instance side

packages/MarkdownEditor-Core.package/MarkdownCommentBlock.class/instance/commentEnd.st
```diff
@@ -0,0 +1,8 @@
+accessing
+commentEnd
+	
+	| closingTag |
+	closingTag := self closingTag.
```
pixunil: As the message is short enough, the variable could be removed in favour of writing `self closingTag`.

pixunil (approved):
No big suggestions left. But the discussion about the test strings remains...

PR #26, opened 2019-06-26
Removes font size glitch

### LINKED ISSUE #6: Font size will bump while typing
### Steps to reproduce
1. Type `# H1` to achieve a heading with big font size.
2. Use the arrow keys <kbd>←</kbd> followed by <kbd>→</kbd> to infect the state.
3. Type in a new line a paragraph.
    _You should notice that the characters are typed big and will shrink after a short time to their normal size._

### Research
The `SmalltalkEditor`, which is used as default for text morphs saves its last seen emphasis into `emphasisHere`. Through the toggling, this variable will contain the emphasis for a heading and won't reset when typing the paragraph.

felix-gohla:
- Fixes #6
- Introduces `UnextendableTextFontReference` which tells the editor window to not use the current font size as next guess. Instead, the standard font is used until the Markdown stylers did finish their work.

I think that it is very hard to write a proper test for this. But testing manually showed, that I cannot reproduce the defect from #6 anymore.

Do you have a better name for `UnextendableTextFontReference`?

### DIFF
--- a/packages/MarkdownEditor-Tests.package/UnextendableTextFontReferenceTest.class/instance/testMayNotBeExtended.st
+++ b/packages/MarkdownEditor-Tests.package/UnextendableTextFontReferenceTest.class/instance/testMayNotBeExtended.st
@@ -1,0 +1,6 @@
+tests
+testMayNotBeExtended
+
+	| fontReference |
+	fontReference :=UnextendableTextFontReference new.
+	self deny: fontReference mayBeExtended
--- a/packages/MarkdownEditor-Tests.package/UnextendableTextFontReferenceTest.class/instance/testMayNotBeExtended.st
+++ b/packages/MarkdownEditor-Tests.package/UnextendableTextFontReferenceTest.class/instance/testMayNotBeExtended.st
@@ -2,5 +2,5 @@ tests
 testMayNotBeExtended
 
 	| fontReference |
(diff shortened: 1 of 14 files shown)


packages/MarkdownEditor-Tests.package/UnextendableTextFontReferenceTest.class/instance/testMayNotBeExtended.st
```diff
@@ -0,0 +1,6 @@
+tests
+testMayNotBeExtended
+
+	| fontReference |
+	fontReference :=UnextendableTextFontReference new.
```
jonaskordt: missing space

jonaskordt (approved):
I think the test is good enough. As for the Name. We could of course just call it `MarkdownTextFontReference` but I'm not sure if that is better.

lpfennigschmidt (approved):
Looks good. I think the name is great, it says everything that you need to know :D

kiragrammel (approved):
great 👍🏼

pixunil (approved):
I tried to write a test which emulates user behaviour (first adding a heading, then a second line and checking whether there are attributes concerning the font size in the second line), but I guess that testing that `UnextendableTextFontReference` does implement the interface correctly and testing that we use this class and not the default, it should cover the glitch...

PR #27, opened 2019-06-28
Refactors tokenization in InlineTextStyler

pixunil:
- [x] Tokens are created character by character
- [x] MarkdownString now contains multiple characters
- [x] Adds postprocessing pass to inform the tokens about their bordering characters
- [x] Removes `MarkdownReadStream`
- [x] Make better naming
- [x] Decide whether `locatedBetween:and:` should be a `subclassResponsibility`
- [x] Remove `tokenClasses` instance variable


packages/MarkdownEditor-Core.package/MarkdownEmphasisDelimiter.class/instance/decideLeftFlankingForPreceding.following..st
```diff
@@ -0,0 +1,9 @@
+initialize-release
+decideLeftFlankingForPreceding: precedingCharacter following: followingCharacter
+
+	| followedByWhitespace followedByPunctuation precededByWhitespaceOrPunctuation |
+	followedByWhitespace := self whitespaceCharacters includes: followingCharacter. 
+	followedByPunctuation := self punctuationCharacters includes: followingCharacter.
+	precededByWhitespaceOrPunctuation := self punctuationCharacters, self whitespaceCharacters includes: precedingCharacter.
+	self isLeftFlanking: (followedByWhitespace not
+		and: [followedByPunctuation ==> precededByWhitespaceOrPunctuation])
```
jonaskordt, line 9: Well done on these methods. They got much better.

packages/MarkdownEditor-Core.package/MarkdownInlineTextStyler.class/instance/initialize.st
```diff
@@ -4,5 +4,4 @@ initialize
 	super initialize.
 	self 
 		clean;
-		tokenClasses: {MarkdownEmphasisDelimiter. MarkdownString} asSortedCollection.
-	self tokenClasses sortBlock: [:a :b | a precedence > b precedence]
\ No newline at end of file
+		tokenClasses: {MarkdownEmphasisDelimiter. MarkdownString}
```
jonaskordt: There is really no reason to have this as an instance variable. Either the accessor could be hardcoded or since the accessor is only used once you could have the collection in the code there. 
(We also did this in the comment block PR since we wanted to make it consistent to the Block selection in the BlockTextStyler so I'm looking forward to merge conflicts :D)

packages/MarkdownEditor-Core.package/MarkdownInlineTextStyler.class/instance/postprocessTokens.st
```diff
@@ -0,0 +1,7 @@
+parsing
+postprocessTokens
```
jonaskordt: I'm not happy with the name. `postprocess` sounds like it is something that happens after everything is done. Doesn't make sense to me that we `interpret` after postprocessing. Maybe just call it `processAdjacentTokens`?

packages/MarkdownEditor-Core.package/MarkdownToken.class/instance/extendWith..st
```diff
@@ -0,0 +1,4 @@
+extending
+extendWith: aCharacter
```
jonaskordt: Weird method name. I would like `append:` better.

I know you are trying to make it consistent with `isExtendableWith:` but I'd like to stick to convention and call it append.

packages/MarkdownEditor-Core.package/MarkdownToken.class/instance/locatedBetween.and..st
```diff
@@ -0,0 +1,4 @@
+initialize-release
+locatedBetween: precedingCharacter and: followingCharacter
+
+	^ self
```
jonaskordt: I think this is weird too. It you just want to return self then you could just leave it empty. I think a `self subclassResponsibility` would make more sense even.

jonaskordt (changes requested):

packages/MarkdownEditor-Core.package/MarkdownToken.class/instance/locatedBetween.and..st
pixunil: I convinced Felix that this is a idiom in Smalltalk to show "We intentionally left this method blank". Yes, we could make this a `subclassResponsibility` and move the empty method into `String` when it makes things clearer

jonaskordt: Tow questions:
- If this is the idiom why don’t we use it in reset?
- Isn’t it better to not provide a default implementation if that default implementation is doing nothing like it is here?

pixunil: - We propose do to so in #21 
 https://github.com/hpi-swa-teaching/MarkdownEditor/blob/7146c85492af2de634ff5f5ec1f6f305f38e429b/packages/MarkdownEditor-Core.package/MarkdownBlockTextStyler.class/instance/reset.st#L2-L4
- Actually, this method is a message in a communication way: we just inform the token about its adjacent characters. Thus, I think a simple "ok, nothing to do" is a sane default. As I said, I don't see a problem with `subclassResponsibility` either.

packages/MarkdownEditor-Core.package/MarkdownToken.class/class/newFor.at..st
```diff
@@ -0,0 +1,7 @@
+instance creation
+newFor: aString at: startPosition
+
+	^ self new 
+		content: aString;
+		startPosition: startPosition;
+		yourself
```
lpfennigschmidt: tab after 'yourself'

packages/MarkdownEditor-Core.package/MarkdownEmphasisDelimiter.class/instance/locatedBetween.and..st
```diff
@@ -0,0 +1,6 @@
+initialize-release
+locatedBetween: precedingCharacter and: followingCharacter
```
lpfennigschmidt: I think this method name should state what the method does. Especially down in 'postProcessTokens', the method call is confusing. You say 'token do something' and then it's just 'token locatedBetween', which does not say what the token actually _does_ with itself. It is setting the flanking for a token, so a suggestion would be '(decide)flankingForPreceding:Following:' and then that splits into deciding left and right flanking.

lpfennigschmidt (changes requested):
Very nicely done, I just have two comments :D

packages/MarkdownEditor-Core.package/MarkdownEmphasisDelimiter.class/instance/locatedBetween.and..st
pixunil: As a general token has no concept of flanking, I'd something that is not specific to flanking, like `processPreceding:following:` (or other)

jonaskordt: What about `setPreceding:following`? I mean we are informing the token about its preceding and following characters so set fits it pretty well. The fact that it doesn't save them in an instance variable isn't really a problem. The set-method idiom still applies.

packages/MarkdownEditor-Core.package/MarkdownInlineTextStyler.class/instance/initialize.st
pixunil: We could even do (which encodes the catch-all behaviour better)
```smalltalk
tokenClass := (MarkdownEmphasisDelimiter mayBeginWith: aCharacter)
	ifTrue: [MarkdownEmphasisDelimiter]
	ifFalse: [MarkdownString].
```

That way, `mayBeginWith:` does not need to be implemented in `MarkdownString` and can be removed as `subclassResponsibility`

jonaskordt: Yea I guess since we only have two kinds of tokens this makes sense. My thinking was to make it as consistent with the BlockStyler as possible, but simple design dictates your solution.

pixunil: If consistency is important to you, I'm also fine with
```smalltalk
matchingTokenClass := {MarkdownEmphasisDelimiter}
	detect: [:tokenClass | tokenClass mayBeginWith: aCharacter]
	ifNone: [MarkdownString].
```
It's a bit more code and has a single-element collection, but it is the same pattern like in block parsing. Which would you take now?

jonaskordt: I like the consistency better.

packages/MarkdownEditor-Core.package/MarkdownToken.class/class/newFor.at..st
pixunil: That's strange - I can see it in Git, but not in Squeak... I'll remove it either way

jonaskordt (approved):

kiragrammel (approved):
Great job! 👍🏼

lpfennigschmidt (approved):

PR #30, opened 2019-07-03
Fixes colouring of instance and global variables in code blocks

felix-gohla:
Adding a `WeakSmalltalkParser` fixes the issue of unknown variables being coloured in the warning color for unknown symbols. It overwrites the checks whether the symbols are known to force them to be coloured normally as if they would exist in the code block.

Please have a look at the tests and whether they are sufficient.


jonaskordt (approved):

pixunil (approved):
It's a bit hidden that the fix is that tokens are interpreted as `#instVar` and not as `#undefinedIdentifier`. But I guess the test is correct.
Maybe we should note it somewhere, but I don't have a constructive idea for it

packages/MarkdownEditor-Tests.package/WeakSmalltalkParserTest.class/instance/testUnkownUppercaseSymbolIsInstanceVariable.st
```diff
@@ -0,0 +1,9 @@
+tests
+testUnkownUppercaseSymbolIsInstanceVariable
```
lpfennigschmidt: should be 'IsGlobalVariable'

packages/MarkdownEditor-Tests.package/WeakSmalltalkParserTest.class/instance/testUnkownLowecaseSymbolIsInstanceVariable.st
```diff
@@ -0,0 +1,9 @@
+tests
+testUnkownLowecaseSymbolIsInstanceVariable
```
lpfennigschmidt: should be 'LowercaseSymbol'
And maybe we can find a better word for symbol (or you can convince me to keep it that way), but the parser doesn't think in symbols, actually they use the word 'token' in their code, or they start comparing 'strings'. Maybe even something like 'unknownVariableIsTreatedLikeInstanceVariable' or something.

lpfennigschmidt (approved):

packages/MarkdownEditor-Tests.package/WeakSmalltalkParserTest.class/instance/testUnkownLowecaseSymbolIsInstanceVariable.st
felix-gohla: Often they use keyword. Would this be okay, too?

packages/MarkdownEditor-Tests.package/WeakSmalltalkParserTest.class/instance/testUnkownUppercaseSymbolIsInstanceVariable.st
felix-gohla: Fixed

felix-gohla:
> It's a bit hidden that the fix is that tokens are interpreted as `#instVar` and not as `#undefinedIdentifier`. But I guess the test is correct.
> Maybe we should note it somewhere, but I don't have a constructive idea for it

Or we add a `self deny: #undefinedIdentifier equals: parsedRanges last type`?

packages/MarkdownEditor-Tests.package/WeakSmalltalkParserTest.class/instance/testUnkownLowecaseSymbolIsInstanceVariable.st
lpfennigschmidt: Yep, that sounds good. But I'm still kinda stumbling over the 'isInstanceVariable'. Because we don't actually care whether it is one or not, we just want it looking like one. Haha, maybe 'looksLikeInstanceVariable' xD No I don't know, but do you get what I mean?

felix-gohla: It is parsed as such. This is what we check. We don't check whether it looks like it.

kiragrammel (approved):
👍🏼

pixunil:
> > It's a bit hidden that the fix is that tokens are interpreted as `#instVar` and not as `#undefinedIdentifier`. But I guess the test is correct.
> > Maybe we should note it somewhere, but I don't have a constructive idea for it
> 
> Or we add a `self deny: #undefinedIdentifier equals: parsedRanges last type`?

The `assert` should be positive, and I would find it more strange that we `assert` that it is `#instVar` and `deny` that it is `#undefinedIdentifier` at the same time, so I'd keep it as it is

PR #31, opened 2019-07-04
Adds contributing

pixunil:
- Includes updated code style
- Includes common pitfalls

Best viewed by using the "rich diff" option.


jonaskordt (approved):
Wonderful 🚟🚃🚋🚞🚝🚄🚅🚈🚂🚆🚇🚊🚉

kiragrammel (approved):
☺️👍🏼🚃

lpfennigschmidt (approved):
I like it, but maybe we can discuss in a group, whether we missed something or not :D

felix-gohla (changes requested):
Maybe we could add an example for block formatting (i.e. block arguments and block instance variables)? :+1:
Do remaining user stories belong into this file and do we even want to include those into the repo?

felix-gohla (approved):
Great 😊

PR #33, opened 2019-07-08
Introduction of Line Object

jonaskordt:


packages/MarkdownEditor-Core.package/MarkdownLine.class/instance/isIndented.st
```diff
@@ -0,0 +1,8 @@
+testing
+isIndented
+
+	| expandedTabsContent |
```
lpfennigschmidt: maybe `contentWithExpandedTabs`? :D

packages/MarkdownEditor-Tests.package/MarkdownCodeBlockTest.class/instance/testFenceLength3To6.st
```diff
@@ -0,0 +1,12 @@
+tests
+testFenceLength3To6
+
+	(Dictionary newFrom: {'```'->3. '```smalltalk'->3.
```
lpfennigschmidt: I would like to leave a newline before the dictionary starts, that way we would have all entries nicely in one block, visually

packages/MarkdownEditor-Tests.package/MarkdownLineTest.class/instance/test0To3LeadingSpacesWithTabAreIndented.st
```diff
@@ -0,0 +1,9 @@
+tests - indentation
+test0To3LeadingSpacesWithTabAreIndented
+
+	{'', Character tab, '1 tab'.
```
lpfennigschmidt, line 4: also make these align on the left :D

packages/MarkdownEditor-Tests.package/MarkdownLineTest.class/instance/testPotentialBlockClass.st
```diff
@@ -0,0 +1,11 @@
+tests - block conversion
+testPotentialBlockClass
```
lpfennigschmidt: `testPotentialBlockClasses`, we're not naming tests after their method name anymore ;)

packages/MarkdownEditor-Tests.package/MarkdownStringTest.class/instance/testIsExtendableWithAnyCharacaterExceptAsterisk.st
```diff
@@ -1,7 +1,7 @@
 tests - recognition
-testIsExtendableWithAnyCharacaterButAsterisk
+testIsExtendableWithAnyCharacaterExceptAsterisk
```
lpfennigschmidt: `testIsExtendableWithAnyCharacterExceptAsterisk`

lpfennigschmidt (changes requested):
:D

packages/MarkdownEditor-Tests.package/MarkdownLineTest.class/instance/testPotentialBlockClass.st
jonaskordt: What would be a better name? I don't really want to list alle the Block Types that we are testing in the test name.

kiragrammel (approved):
It is a pity that in some methods more lines have been added, but it has become more beautiful in other methods. 👍🏼

packages/MarkdownEditor-Tests.package/MarkdownLineTest.class/instance/testPotentialBlockClass.st
pixunil: The difference between current and proposed name is two characters small ^^

jonaskordt: Ahh. I'm blind I guess

packages/MarkdownEditor-Core.package/MarkdownLine.class/instance/potentialBlockClass.st
```diff
@@ -0,0 +1,6 @@
+accessing
+potentialBlockClass
```
felix-gohla: I don't like the name. When I first read it, I had no idea what it meant. Maybe `potentialMatchingBlockClass`. This expresses that it tries to find a matching class. But maybe it is a little bit to fastidious (I don't have a clue what's the right word for penibel: https://www.dict.cc/?s=penibel 😄)

felix-gohla (approved):
Looks good to me. Only one small comment.

packages/MarkdownEditor-Core.package/MarkdownLine.class/instance/potentialBlockClass.st
jonaskordt: So my thought process was that the line is potentially converted into a block of the potential block class. You are right that we want the class which matches the line. But we have renamed `Block>>matches:` to `mayBeginWith:` to be consistent with the tokens. I think it would be weird to now bring the word match back into the name because we don't use it like that. I would also be okay with just calling it `blockClass` since the matching (which we don't call matching..) is definite and not really potential, only the conversion is potential. Would that be clearer to you?

felix-gohla: Yep, that would help.

lpfennigschmidt (approved):

PR #35, opened 2019-07-10
Simplifies TextAttributes

### LINKED ISSUE #18: Simplify attributes
`MarkdownAttribute` has support for a collection of `TextAttribute`, where we only will use one per object. Thus, it should become a single value.

Also, it has two subclasses which only differ in their value. This should be refactored into a variable in a single class.

felix-gohla:
Fixes #18.

One comment: In the tests we e.g. check against `MarkdownAttribute bold textAttribute`. Should we change that to `TextEmphasis bold`?

Co-authored-by: Kira Grammel <kira.grammel@student.hpi.de>


packages/MarkdownEditor-Core.package/MarkdownAttribute.class/properties.json
```diff
@@ -4,10 +4,11 @@
...
 	"classvars" : [
 		 ],
-	"commentStamp" : "fgo 5/21/2019 15:40",
+	"commentStamp" : "fgo 7/10/2019 15:23",
 	"instvars" : [
 		"startPosition",
-		"endPosition" ],
+		"endPosition",
+		"textAttribute" ],
 	"name" : "MarkdownAttribute",
```
pixunil: I am sorry for the late suggestion, but actually, the `MarkdownAttribute` would be better described as `MarkdownEmphasis` which provides `normal` (old `italic`) and `strong` (old `bold`). That way, the mapping between the Markdown domain and the Squeak implementation would be more obvious and there would be no `attribute textAttribute` naming duplication.

But if you say that renaming this class isn't worth it, I'm fine with it

packages/MarkdownEditor-Core.package/MarkdownAttribute.class/README.md
```diff
@@ -1 +1,2 @@
-I am an attribute for a text and have a range. StartPosition and endPosition are counted inclusively.
\ No newline at end of file
+I am an attribute for a text and have a range. StartPosition and endPosition are counted inclusively.
```
pixunil:
```suggestion
I represent a Markdown emphasis over a range of characters. The applied styling is expressed through my textAttribute.
```

packages/MarkdownEditor-Tests.package/MarkdownInlineTextStylerTest.class/instance/testInterpretTokensCutDelimiterClosesTwice.st
```diff
@@ -1,17 +1,20 @@
...
 		interpretTokens.
 	
 	self assert: 2 equals: inlineTextStyler attributes size.
 	self assert: 6 equals: inlineTextStyler tokens size.
 	
-	self assert: (inlineTextStyler attributes first isKindOf: MarkdownAttributeBold).
-	self assert: 3 equals: inlineTextStyler attributes first startPosition.
-	self assert: 7 equals: inlineTextStyler attributes first endPosition.
+	firstAttribute := inlineTextStyler attributes first.
+	self assert: MarkdownAttribute bold textAttribute equals: firstAttribute textAttribute.
```
pixunil: Actually, we want to test the attribute / emphasis structure similar to the block structure here. I would propose to generalize this a bit, like:

```smalltalk
self
    assert: {MarkdownAttribute bold setFrom: 3 to: 7. MarkdownAttribute italic setFrom: 1 to: 8}
    areEmphasesIn: inlineTextStyler attributes
```
I know that this proposal must be modified because `setFrom:to:` is not quite correct in this context and we need to implement the comparing logic in a new utility, but I think that a similar approach would greatly improve the tests.

pixunil (changes requested):
Some ideas come late, but I hope not too late ^^'

packages/MarkdownEditor-Core.package/MarkdownAttribute.class/properties.json
felix-gohla: But I think `normal` would conflict with `TextEmphasis normal` which is - well - normal text...
But we'll rename `MarkdownAttribute` to `MarkdownEmphasis`.

jonaskordt: Maybe we could call them `weak` and `strong`. I would really like to create the mapping of spec to Smalltalk like Jakob suggested.

lpfennigschmidt: If you rename it, could we also change the local variables in the test cases that are now called `attribute` to `emphasis`? If you don't want to do it, I can also do it, but then also the access is nicer with `emphasis textAttribute` instead of `attribute textAttribute`

felix-gohla: What about `MarkdownEmphasis from:to:` being italic and `MarkdownEmphasis strongFrom: to:` being bold?

lpfennigschmidt (dismissed):
Like it, but I also agree with the others about renaming to MarkdownEmphasis and adding a utility method for testing :D

packages/MarkdownEditor-Core.package/MarkdownAttribute.class/properties.json
felix-gohla: > If you rename it, could we also change the local variables in the test cases that are now called `attribute` to `emphasis`? If you don't want to do it, I can also do it, but then also the access is nicer with `emphasis textAttribute` instead of `attribute textAttribute`

Yeah, we could do `emphasis textAttribute`. No problem.

jonaskordt: > What about `MarkdownEmphasis from:to:` being italic and `MarkdownEmphasis strongFrom: to:` being bold?

That sounds good.

felix-gohla: Done

packages/MarkdownEditor-Core.package/MarkdownAttribute.class/README.md
felix-gohla: Done

packages/MarkdownEditor-Core.package/MarkdownEmphasisDelimiter.class/instance/attributeFrom.to.delimiterLength..st
```diff
@@ -0,0 +1,6 @@
+attributes
+attributeFrom: aStartPosition to: anEndPosition delimiterLength: aLength
```
pixunil:
```suggestion
attributeFrom: startPosition to: endPosition delimiterLength: aLength
```
We do variable shadowing elsewhere. If this bothers you, I'd prefer `attributeStartPosition`

packages/MarkdownEditor-Core.package/MarkdownEmphasisDelimiter.class/instance/attributeStartingAt..st
```diff
@@ -1,9 +1,9 @@
...
 
-	| attributeDelimiterLength start end attributeClass |
-	attributeDelimiterLength := (self delimiterLengthForAttributeWith: openingMarkdownEmphasisDelimiter).
-	start := openingMarkdownEmphasisDelimiter endPosition - attributeDelimiterLength + 1.
-	end := self startPosition + attributeDelimiterLength - 1.
-	attributeClass := self attributeTypeFor: attributeDelimiterLength.
-	^ attributeClass from: start to: end
+	| delimiterLength start end |
+	delimiterLength := self delimiterLengthForAttributeWith: openingEmphasisDelimiter.
+	start := openingEmphasisDelimiter endPosition - delimiterLength + 1.
```
pixunil: I'd rename these inline variables if the line does not get too long

packages/MarkdownEditor-Tests.package/MarkdownEmphasisDelimiterTest.class/instance/testDifferentDelimitersCreateBoldAttribute.st
```diff
@@ -1,17 +1,11 @@
...
+	opener := MarkdownEmphasisDelimiter newFor: '****' at: 1.
+	closer := MarkdownEmphasisDelimiter newFor: '***' at: 6.
 					
-	attribute := closer attributeStartingAt: opener. 	
-	self assert: (attribute isKindOf: MarkdownAttributeBold).
-	self assert: 3 equals: attribute startPosition.
-	self assert: 7 equals: attribute endPosition
\ No newline at end of file
+	emphasis := closer attributeStartingAt: opener. 	
+	self assert: TextEmphasis bold equals: emphasis textAttribute.
```
pixunil: You could compare `emphasis` with `MarkdownEmphasis strongFrom: 3 to: 7`.

Applies also to `testEmphasisAttributeItalic`.

packages/MarkdownEditor-Tests.package/MarkdownEmphasisDelimiterTest.class/instance/testDifferentDelimitersCreateBoldAttribute.st
```diff
@@ -1,17 +1,11 @@
-tests - attributes
+tests - emphases
 testDifferentDelimitersCreateBoldAttribute
```
pixunil: `testDifferentDelimitersCreateStrongEmphasis`

packages/MarkdownEditor-Tests.package/MarkdownEmphasisDelimiterTest.class/instance/testSameDelimitersCreateBoldAttribute.st
```diff
@@ -1,17 +1,11 @@
-tests - attributes
+tests - emphases
 testSameDelimitersCreateBoldAttribute
```
pixunil: `testSameDelimitersCreateStrongEmphasis`

packages/MarkdownEditor-Tests.package/MarkdownInlineTextStylerTest.class/instance/testDelimiterCantCloseItself.st
```diff
@@ -20,6 +20,7 @@ testDelimiterCantCloseItself
 				(MarkdownString new content: 'b'; yourself)}.
 	
 	inlineTextStyler findOpenerFor: openCloser 
-		ifFound: [:o | foundOpener := o].
-	self deny: (openCloser = foundOpener).
-	self assert: opener equals: foundOpener
\ No newline at end of file
+		ifFound: [:foundOpener | 
+			self deny: (openCloser = foundOpener).
```
pixunil, line 24: Isn't this `deny` redundant? We know that `opener` and `openCloser` are two variables, and when we assert that the `foundOpener` is `opener`, it can't be `openCloser` by transitivity, right?

packages/MarkdownEditor-Tests.package/MarkdownInlineTextStylerTest.class/instance/testInterpretTokensAsItalics.st
```diff
@@ -1,11 +1,11 @@
 tests - interpretation
 testInterpretTokensAsItalics
 
+	| expectedEmphases |
```
pixunil: I don't see the advantage of an inline variable instead of inlining the array and putting some breaks in the assert call.

pixunil (changes requested):
Now I don't fear those inline tests anymore ^^
Shouldn't `attributes` now become `emphases`?
Regarding the `assert:areEmphasesIn:`, I find it strange to supply `hash`. Supplying `=` is not ideal, but is relatively trivial, whereas `hash` looks really like having testing logic where it does not belong

jonaskordt (changes requested):
I agree with Jakob mostly. 

The main problem I see is having the `MarkdownEmphasis>>=` and `MarkdownEmphasis>>hash` in production code, where it doesn't belong. 

The rest is fine with me.

packages/MarkdownEditor-Core.package/MarkdownEmphasisDelimiter.class/instance/attributeFrom.to.delimiterLength..st
felix-gohla: I don't like `attributeStartPosition`. We know that everything we do in this method is regarding attributes. Thus, this would be redundant. Yeah, we tried to fix the shadowing elsewhere, too.

packages/MarkdownEditor-Core.package/MarkdownEmphasisDelimiter.class/instance/attributeStartingAt..st
felix-gohla: To what? The line lengths would be okay, I guess.

packages/MarkdownEditor-Tests.package/MarkdownEmphasisDelimiterTest.class/instance/testDifferentDelimitersCreateBoldAttribute.st
felix-gohla: Kira and me, we discussed this. In my opinion, it is another layer of abstraction because this comparison more feels like integration tests whereas this is a unit test.

packages/MarkdownEditor-Tests.package/MarkdownInlineTextStylerTest.class/instance/testDelimiterCantCloseItself.st
felix-gohla: We just left it in. In my point of view, this "redundancy" makes the test easier to understand. From a logical standpoint, you're absolutely right. But pointing out, that it is not the `openCloser` but the `opener` adds some clarity.

packages/MarkdownEditor-Core.package/MarkdownEmphasisDelimiter.class/instance/attributeStartingAt..st
pixunil: I wanted to rename them `startPosition` or `endPosition`, but after the discussion about the arguments `attributeFrom:to:`, I'd keep them here and rename the arguments of `attributeFrom: start to: end`

packages/MarkdownEditor-Tests.package/MarkdownInlineTextStylerTest.class/instance/testDelimiterCantCloseItself.st
pixunil: Okay, in my case it was the opposite, because I have both of a positive and a negative assert and we actually want to test equality. As a reader, I need to decide whether the deny or the assert is the important part. Thus, it doesn't make it clearer for me, but confuses me...

packages/MarkdownEditor-Tests.package/MarkdownEmphasisDelimiterTest.class/instance/testDifferentDelimitersCreateBoldAttribute.st
pixunil: I know that `BoldEmphasis` matches `ItalicEmphasis`, but there is only `StrongEmphasis` in Markdown-talk.
How about `StrongEmphasis` and `NormalEmphasis` / `SimpleEmphasis`? (alternatives welcome)

pixunil (approved):
Do as much as you want to do of the open issues, but I think the core idea of this refactoring is implemented :D

jonaskordt (approved):

packages/MarkdownEditor-Tests.package/MarkdownEmphasisDelimiterTest.class/instance/testDifferentDelimitersCreateBoldAttribute.st
pixunil: As discussed with SWA people, both variants are fine. If the comparison gets repeated often (becomes a pattern), a shorter way is preferred (custom assert or just using `=`).
As this happens only a handful of time, I'm indifferent about this

Issue #34, opened 2019-07-10
Class comments are missing


felix-gohla:
... in most of the classes. 😕

PR #36, opened 2019-07-10
Deduplicates test conveniences

### LINKED ISSUE #28: Code duplication in TestCase accessors
Some accessors in the `TestCase` subclasses are used by multiple subclasses. This produces code duplication.

Solution: Create a `TestCase` subclass which implements those accessors and let the other subclasses inherit that implementation.

pixunil:
Resolves #28 

Also does some work for #34. We removed some class comments here as they didn't provide any value.

This PR will wait for #33, as we can remove `MarkdownBlockTest` afterwards.


kiragrammel (approved):
👍🏼

jonaskordt (approved):

felix-gohla (approved):
Great 😊

PR #133, opened 2021-05-26
Feature/alternative delimiters

### LINKED ISSUE #124: Alternate delimiters for bold/italic
As a **User** I want to use the **"_" Syntax to hightlight bold and italic** text instead of the *

### LINKED ISSUE #134: HTML Export for nested inlineelements
When having multiple Inlineelements nested in eachother the HTML export fails to export all of them.
This Problem occurs in `MarkdownHtmlConverter -> createNestedHTML: aNested`

JulianEgbert:
Implement _ Delimiter for __Bold__ and _Italic_ in addition to *-Delimiter as requested in #124.


JulianEgbert:
Html Converter Fails for nested elements as explained in #134

PR #151, opened 2021-06-03
Fix Coverage by excluding methods

JulianEgbert:
The methods to exclude have to be in a specific categorie. This is currently called `'open / save - I/O'`.


JulianEgbert:
Not a solution for the problem.

PR #160, opened 2021-06-24
merge inline parsing into new parser

ghost:
inline parsing autolinks, strong, italic, strike emphasis, alternative emphasis delimiter, normal links and images (aka references)
punctuation character rule
divisible by three rule
special rule for alternative delimiters

inline parsing starts at MarkdownInlineParser parse


ghost:
included in pull request 161

